Issue #26, opened 2022-07-13
Respect "gameover" phase


benjaminpjones:
If the game is over, we should show the result and prevent new moves from being submitted to the server.

benjaminpjones:
d22ac69a0fdc325037033e631d370b3ae3ed9d9a shows the result. We still need to prevent further moves from getting played.

benjaminpjones:
Re-opening - it looks like this didn't make it through the Vue rewrite.

Issue #25, opened 2022-07-13
default configs


benjaminpjones:
We should have default configs for each game, since right now it defaults to a config that works for baduk.

PR #27, opened 2022-07-13
Add TicTacToe (aka naughts and crosses)

benjaminpjones:
This is a demo of how to add a game (tic-tac-toe) to the repo.

Screen Shot 2022-07-12 at 9 30 49 PM
Screen Shot 2022-07-12 at 9 27 02 PM

### DIFF
--- a/packages/client/src/tictactoe/TicTacToeView.tsx
+++ b/packages/client/src/tictactoe/TicTacToeView.tsx
@@ -1,0 +1,31 @@
+import { GameViewProps } from "../view_types";
+import { map2d, GridBoard, Stone } from "../boards";
+import { TicTacToeState } from "@ogfcommunity/variants-shared";
+
+export function TicTacToeView({
+  gamestate,
+  onMove,
+}: GameViewProps<TicTacToeState>) {
+  // We could implement a tic tac toe board, but since this is a go variants
+  // server, might as well use a Go board :)
+  // The following statement turns Xs and Os into something GridBoard can
+  // understand
+  const board: Stone[][] = map2d(gamestate.grid, (space_state) => {
+    switch (space_state) {
+      case "x":
+        return { color: "black" };
+      case "o":
+        return { color: "white" };
+      case undefined:
+        return {};
+    }
+  });
+
+  // e.g. if player 0 clicks the center of a 3x3 board,
+  // the move object { 0: "11" } will be sent to the server.
+  const onClick = (pos: { x: number; y: number }) => {
+    onMove({ [gamestate.next_to_play]: pos.x.toString() + pos.y.toString() });
+  };
+
+  return <GridBoard board={board} onClick={onClick} />;
+}
--- a/packages/client/src/tictactoe/index.ts
+++ b/packages/client/src/tictactoe/index.ts
@@ -1,0 +1,1 @@
+export * from "./TicTacToeView";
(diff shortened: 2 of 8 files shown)


benjaminpjones:
Doing this exercise made me realize there are some opportunities for improvement:

- Game objects and game views can be implemented in a single short file.  This means that we may not need to put each game in its own directory. This would reduce the boilerplate (new dir, index.js)
- The end of a game isn't really handled yet (#26)
- Config is hardcoded to something for baduk. We should allow variants to specify a default config (#25)
- We can move grid utils into a library in shared (map2d, make empty grid, fill grid, other 2D analogs to Array functions)
- errors aren't respected yet
   - We should expect games to throw in the constructor (illegal config) and playMove (illegal move)

JonKo314:
I'm wondering whether we should make the code for the variants less redundant. TicTacToe is like Go without captures and a different win condition, so the two variants could share a lot of code. Maybe the only parts that should be implemented for TicTacToe is the *transition function* (where the capturing happens in Go) and the ending of the game.

benjaminpjones:
Yeah I think a transition function would be like the functional version of playMove and exportState combined.  Still, not sure I see how we can get away from writing config validation (currently done in `TicTacToe.constructor`) or number of players/nextToPlay (although we could default to two alternating players if that's what you're getting at)

JonKo314:
> Yeah I think a transition function would be like the functional version of playMove and exportState combined.

I don't follow, the transition function happens after players have made their move and before they are allowed to make new moves. For Baduk it has to remove captured stones, for TicTacToe it maybe just needs check if the game is over.

> Still, not sure I see how we can get away from writing config validation (currently done in `TicTacToe.constructor`) or number of players/nextToPlay

Well, config validation probably needs to be added to AbstractGame (which should maybe be renamed to AbstractVariant). Then we either write a super class for both variants or we load the same code in some different way. Coming up with good names might be the hardest part. :sweat_smile: :wink:

benjaminpjones:
>I don't follow, the transition function happens after players have made their move and before they are allowed to make new moves. For Baduk it has to remove captured stones, for TicTacToe it maybe just needs check if the game is over.

I think if you look at playMove(), that's pretty much all that's in that method..

JonKo314:
I think that's too much in that method then.

benjaminpjones:
I think it might be easier to have a fruitful discussion if you make a PR or commit that changes AbstractGame to the way you think it should be. Then we have something concrete to discuss pros and cons.

I'm happy to change the way things are set up, but it's hard for me to address criticism without a clear alternative.

JonKo314:
Sorry, you are right, I should put my thoughts into code.

JonKo314:
Okay, I've done some refactoring of both variants to figure out how the functionalities they share could be implemented with less redundancy: https://github.com/benjaminpjones/govariants/commit/01c399f982b3ac09000481708fc0b6f4d1b3dbfc

Now we'd have to figure out if aiming for minimal redundancy is worth the trouble, or if it's simpler and sufficient to keep the variants separated.
Also I'm not sure whether this pull request is the "right" place to discuss this stuff.

benjaminpjones:
Closed this PR, and added a section in the README that points to it: https://github.com/benjaminpjones/govariants/blob/main/README.md#creating-a-new-variant.  We can update as the process changes :)

PR #32, opened 2022-07-19
Create a GitHub workflow (CI)

benjaminpjones:
This is so that the tests get run whenever there is a push or PR to main.  In the future we might want to run a linter in this workflow too.


benjaminpjones:
Awesome, the test failure is expected, I'll merge this then fix the build in a sec.

PR #33, opened 2022-07-19
Add jest as test script.

benjaminpjones:
Specifies jest as the test command (at least for `/shared`).  this should fix the fact that the build is broken.

From the root of the repo you can now run `npm test` or `yarn test` and it will run jest for the shared repo, and any other test scripts we add in the future.


benjaminpjones:
Woo a check mark!

PR #41, opened 2022-08-14
User DB abstraction

benjaminpjones:
This takes raw Mongo calls out of index.ts.  In addition, this lays groundwork for handling both guest and persistent users in a similar way.

#### Guest interface

- `interface GuestUser`
- `function getUserBySessionId(id)`
- `function createUserWithSessionId(id)`

#### Persistent interface

- `interface PersistentUser`
- `function getUserByName(username)`
    - Note: this one may be unnecessary, maybe a more useful function would be `authenticateUser(username, password)`
- `function createUserWithUsernameAndPassword(username, password)`

#### Login method agnostic

- `function getUser(id)`
- `function deleteUser(id)`
- `interface UserResponse`

### DIFF
--- a/packages/server/src/games.ts
+++ b/packages/server/src/games.ts
@@ -23,6 +23,10 @@ export async function getGame(id: string): Promise<GameResponse> {
     _id: new ObjectId(id),
   });
 
+  if (!db_game) {
+    return undefined;
+  }
+
   return outwardFacingGame(db_game);
 }
 
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -3,25 +3,25 @@ import session from "express-session";
 import http from "http";
 import { Server } from "socket.io";
 import { getGame, getGames, createGame, playMove } from "./games";
+import {
+  createUserWithSessionId,
+  deleteUser,
+  getUser,
+  getUserBySessionId,
+} from "./users";
 import bodyParser from "body-parser";
 import cors from "cors";
 import path from "path";
-import { connectToDb, getDb } from "./db";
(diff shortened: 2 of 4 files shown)


JonKo314 (commented):
Nice :+1:, just one question:

```
export async function getGame(id: string): Promise<GameResponse> {
  const db_game = await gamesCollection().findOne({
    _id: new ObjectId(id),
  });

  if (!db_game) {
    return undefined;
  }

  return outwardFacingGame(db_game);
}
```

Is there a reason why we don't throw an error instead of returning `undefined`?

benjaminpjones:
Thanks!

>Is there a reason why we don't throw an error instead of returning undefined?

No reason in particular.  Is there a reason to throw instead of returning `undefined`?

I am fine with either, as long as all the database getter functions are consistent - currently they all return `undefined` if there's no matching entry in the database.

JonKo314:
I think it depends on whether we consider calls of the function with unknown ids as valid or not. The funcion could either say "I don't know this id, you must have made a mistake" or "I don't know this id, but I'm sure you know what you're doing, here have nothing". Although looking at how the mongodb package handles things, returning `null` or `undefined` is probably more consistent.

benjaminpjones:
Hmm that's a tough question to answer haha. I guess one evidence that we consider it a "valid" call is the way we call `getUserBySessionId()` - if it fails, we just create the user.  So it's not a programming error (although the following example would look fine as a try/catch block too)

```
let user = getUserBySessionId();
if (!user) {
    user = createUserWithSessionId()
}
```

Issue #36, opened 2022-07-31
Add a lint check


benjaminpjones:
I noticed there's about 11 lint errors in the Vue app, but nothing enforcing it. Here's a list of things that we need in order to get the whole repo linted:

- [x] Add eslintrc file for the packages that don't have it (shared & server?)
    - To start, I think we want to extend `"eslint:recommended"`
- [x] Add a lint script to the root [package.json](https://github.com/benjaminpjones/govariants/blob/main/package.json#L8-L15)
- [x] Fix any lint errors that show up
- [x] Call the lint script in the [GitHub workflow](https://github.com/benjaminpjones/govariants/blob/b382e5a7ffc88f9e7285bbf5f7f6280285baf38f/.github/workflows/node.js.yml)

There are also a variety of IDE integrations available for ESLint that folks can use for faster feedback: https://eslint.org/docs/latest/user-guide/integrations

PR #42, opened 2022-08-15
Turn on lint check (Vue only) in CI

benjaminpjones:
Y'all good with enforcing ESLint?  The config file was already in there, but nothing was enforcing it (besides the red squigglies in my IDE 😁)

This PR addresses some of the tasks in #36 

## Changes

- Add a top level lint command
- Call that command in the GitHub workflow
- Fix lint errors that popped up (Vue only)
   - The reason for starting with the Vue repo is that we already have a `.eslintrc` and also there's also more unmerged changes in `shared` and `server` (`authentication` branch)


JonKo314 (commented):
Looking good to me. :+1:

benjaminpjones:
Thanks for reviewing @JonKo314 !

PR #53, opened 2022-09-03
Multicolor Grid Board

benjaminpjones:
This is a milestone for Simultaneous Fractional Go.  I haven't plugged this in to a Variant yet, but it has been tried out a bit on a new route `/components`

Screen Recording 2022-09-02 at 11 19 46 PM


benjaminpjones:
oops, i should wait for the checks to finished next time. fixed the error!

PR #55, opened 2022-09-05
Add better types to game_map and board_map.

benjaminpjones:
There's still an `any` in there, but now at least we know the basic shape of what's stored in these maps.

Note: the `new() => X` is the way to refer to constructor functions in TypeScript (https://www.typescriptlang.org/docs/handbook/2/generics.html#using-class-types-in-generics)


packages/shared/src/game_map.ts
```diff
@@ -2,9 +2,9 @@ import { Baduk } from "./baduk";
 import { AbstractGame } from "./abstract_game";
 import { Phantom } from "./phantom";
 
-// Is there any way we can get away from `any` here?
-// I think only the infra needs to touch these functions, but still!
-export const game_map: { [variant: string]: any } = {
+export const game_map: {
+  [variant: string]: new (config: any) => AbstractGame;
```
merowin, line 6: This syntax is new to me. Does this mean that what comes after the colon must have a constructor that takes a config (of any type) and returns an AbstractGame?

packages/shared/src/game_map.ts
```diff
@@ -18,8 +18,8 @@ class ConfigError extends Error {
 
 export function makeGameObject(
   variant: string,
-  config: any
-): AbstractGame<any, any> {
+  config: unknown
+): AbstractGame<unknown, unknown> {
```
merowin, line 22: I had to look up unknown, but not sure if I understand. Would this create an error before compiling? (I assume so, but not sure)

const config: BadukConfig = { width: 9, height: 9, komi: 6 }
const game: AbstractGame<BadukConfig, BadukState> = makeGameObject("baduk", config)

merowin (approved):

packages/shared/src/game_map.ts
benjaminpjones: There's a lot of colons in here, but yeah I think that's right.

TBH the constructor syntax is new to me too.  I wanted to type this as `{ [variant: string]: AbstractGame }`, but then we run into the problem:

```
return new game_map[variant](config);
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    Cannot create an instance of an abstract class.(2511)
```

packages/shared/src/game_map.ts
benjaminpjones: Exactly!  It basically says "don't make assumptions about what comes out of this function".  There are ways we could type this so that TypeScript knows that `makeGameObject("baduk", config)` always outputs a `Baduk`, but I think that kind of defeats the purpose of this function (we can just call `new Baduk(config)` instead)

You might have noticed I silenced a similar error in vue-client/src/store/baduk.ts using a type assertion:

```
function getStateFromResponse(response: GameResponse): BadukState {
  const game = makeGameObject(response.variant, response.config);
  response.moves.forEach((move) => {
    game.playMove(move);
  });
  return game.exportState() as BadukState;
}
```

I'm not terribly proud of that (as I said earlier - `as` is a way to trick the compiler), but I think the store stuff needs to be modified to handle multiple variants anyway.

benjaminpjones:
Thanks for reviewing!

PR #58, opened 2022-09-08
Locate server ressources at /api for production

JonKo314:
To ensure URL consistency between production and development and to avoid conflicting URLs between client and server projects.


benjaminpjones (approved):
LGTM! Thanks for the improvement!

PR #59, opened 2022-09-09
Parallel Go

benjaminpjones:
I still need to implement captures, but this PR was getting large and it seemed "playable" :) This is our first demonstration of a Simultaneous variant.

Parallel Go demo

You can try Parallel Go with the config: 

```
"config": {
    "width": 19,
    "height": 19,
    "num_players": 4,
    "collision_handling": "merge"
 },
```

Other options for `"collision_handling"` include `"ko"` and `"pass"`.

Resources:
- http://www.di.fc.ul.pt/~jpn/gv/players.htm
- https://senseis.xmp.net/?FGo


benjaminpjones:
I've also picked up a lot of TODOs/side quests along the way:

- Captures should be implemented
- Different color annotations for dark stones, (black on black doesn't show up)
- Clean up `decodeMove`/`encodeMove` (maybe move_utils.ts?)
- Config entry components
- Refactor shared code between `BadukBoard` and `MulticolorGoBoard`.
- Under construction component
- upgrade vue-client target to es2017+ (so I can use Object.entries)
- pass filtered game state via socket instead of full list of moves (one could potentially analyze the JSON and see others' moves)

JonKo314 (approved):
Really cool.

For stone annotation colors we could use css filters or mix-blend-mode. This would also work for multicolored stones, just gray stones might cause some problems (e.g. inverting #808080 doesn't do much).

See:
https://css-tricks.com/methods-contrasting-text-backgrounds/
https://forums.online-go.com/t/fractional-multicolour-go/36996/64

benjaminpjones:
Oh wow yeah I did some searching but hadn't found a property like that. I like that look a lot!

merowin:
This is really cool! Sorry that I didn't have the time to review.

benjaminpjones:
No worries @merowin.  It's live now if you want to try it out! ([example game](https://www.govariants.com/game/631e92db6b9b9b4a0354f551))

PR #61, opened 2022-09-11
Coordinate

benjaminpjones:
There was a lot of duplicate `decodeMove` `encodeMove` `coordsToLetters` (mostly my fault haha).  This PR moves this logic into a class called `Coordinate`.  One alternative of course is to keep these as functions that operate on `CoordinateLike` objects, but to me these functions seem very tied to the {x, y} data structure, hence the object oriented approach - let me know what you think.


benjaminpjones:
Note to reviewers: I think I'm mainly looking for thoughts on the interface.  In other words, is this a reasonable way to switch between `"cd"` and `{ x: 2; y:3 }`?

```
const pos = new Coordinate(2, 3);
const pos2 = Coordinate.fromSgfRepr("cd");
const move_letters = pos.toSgfRepr();
```

PR #62, opened 2022-09-11
Delete the baduk store

benjaminpjones:
So I think this is no longer used.  At the beginning, this was how we handled the API response, but I bypassed that when generalizing to multiple variants.


benjaminpjones:
@JonKo314 Here, looking for confirmation that this isn't actually used and there are no immediate plans to generalize to multiple variants and use it again.

PR #63, opened 2022-09-11
Remove Vue example files

benjaminpjones:
This is basically stuff that was on the "Welcome" page and the "Hello World" examples.  No longer relevant to our app.


benjaminpjones:
This one seems pretty safe, just want to confirm there's nothing in here you want to keep!

JonKo314 (approved):
I might want to look at how the example of testing the user interface is done, but all those files are auto-generated and thus can easily be regenerated to take a look.

benjaminpjones:
Makes sense - automated testing would definitely be good.  But yeah if you think it's easy enough to regenerate (and also they can be looked up in source control) then I'll just take them out for now (esp since `HelloWorld.test` won't make much sense without `HelloWorld.vue`)

Thanks for  reviewing!

PR #64, opened 2022-09-11
Extension recommmnedations

benjaminpjones:
@JonKo314 @merowin We had some discussion in the chat about recommended extensions - I believe this config file tells VSCode to recommend these extensions to people opening this workspace.


benjaminpjones:
Oops the 300 lines is a different PR let me fix that.

PR #69, opened 2022-09-25
Use real users for seats

benjaminpjones:
No longer **One True User**.  To take a seat at a game, you'll need to click the Remember Me button.

Screen Recording 2022-09-24 at 10 28 50 PM

Note:  I haven't included a db migration as part of this PR.  I figure it doesn't cause too much trouble for the One True User to be seated at old games.  At some point we probably want to implement "admin can kick players" anyway.


benjaminpjones:
Review requested from @JonKo314 because you have the most familiarity with the user system.

PR #70, opened 2022-09-26
Add default configs

### LINKED ISSUE #25: default configs

benjaminpjones:
Fixes #25
Related issue: #8 

This makes it so that the initial config is actually valid for the selected variant.

Screen Recording 2022-09-25 at 9 44 36 PM

### DIFF
--- a/packages/shared/src/abstractAlternatingOnGrid/abstractAlternatingOnGrid.ts
+++ b/packages/shared/src/abstractAlternatingOnGrid/abstractAlternatingOnGrid.ts
@@ -28,9 +28,9 @@ export abstract class AbstractAlternatingOnGrid<
   protected next_to_play: 0 | 1 = 0;
   protected last_move = "";
 
-  constructor(config: AbstractAlternatingOnGridConfig) {
+  constructor(config?: AbstractAlternatingOnGridConfig) {
     super(config as TConfig);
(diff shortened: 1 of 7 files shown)


benjaminpjones:
Note: I'm a little torn on putting adding this on to the game object.  For one, this requires constructing the game which lead to a chicken-and-egg problem - you need a game object to get the default config, but you needed a config to construct a game.  I get around this by allowing construction without a config.

Alternative considered:

A `default_config_map` could hold variant:config pairs.

- Pros: Easy to query, no chicken-or-egg problem
- Cons: Hard to make TypeScript enforce validity of the config, another place that game authors need to be aware of

JonKo314:
Yeah, I see the problem. Would it help to make `defaultConfig()` a static member?

benjaminpjones:
I'm not sure `static` works with the abstract class/polymorphic system we have going on.

JonKo314:
Ah, yes. Both characteristics, the method being abstract and the base class being generic, seem to make the use of `static` impossible.

And I've thougth about defining the default config via the config type, but I think that leads to another chicken-and-egg problem and I don't know how we could access the default type from an abstract game object.

So I think we should keep it as it is (at least for now, maybe a better solution comes up in the future). I like that it forces developers to provide a default config when they extend the base class.

PR #71, opened 2022-09-28
Emit game info via socket when a move is played.

### LINKED ISSUE #7: Use socket.io for gameplay
If player A and player B are both on the game page, and player A plays a move, player B's board will not update.  We can use socket.io to send moves to clients as they are played.

benjaminpjones:
Fixes #7 

output


benjaminpjones:
Also reimplemented the "SocketTest" component as a refresher:

Screen Shot 2022-09-27 at 10 15 50 PM

PR #75, opened 2022-10-01
Make viewBox of polygonal board fit intersections

JonKo314:
Before:
grafik

After:
grafik

### DIFF
--- a/packages/vue-client/src/components/boards/BadukBoardAbstract.vue
+++ b/packages/vue-client/src/components/boards/BadukBoardAbstract.vue
@@ -4,6 +4,7 @@ import type {
   BadukWithAbstractBoardConfig,
   BadukWithAbstractBoardState,
 } from "@ogfcommunity/variants-shared/src/badukWithAbstractBoard/badukWithAbstractBoard";
+import { computed } from "vue";
 
 const props = defineProps<{
   config: BadukWithAbstractBoardConfig;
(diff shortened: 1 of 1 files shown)


packages/vue-client/src/components/boards/BadukBoardAbstract.vue
```diff
@@ -25,6 +26,18 @@ const emit = defineEmits<{
...
 }
+
+const viewBox = computed(() => {
+  const xPositions = props.gamestate.board.Intersections.map(
+    (i) => i.Position.X
+  );
+  const yPositions = props.gamestate.board.Intersections.map(
+    (i) => i.Position.Y
+  );
+  return `${Math.min(...xPositions) - 1} ${Math.min(...yPositions) - 1} ${
```
merowin, line 37: At the end of constructing the intersections of the polygonal board, all positions are shifted such that Math.min(...xPositions) and Math.min(...yPositions) are 0, 0. 

However I like that this is handled in the view, so the board is always pictured nicely. So should we remove the abovementioned shifting in the construction algorithm?

merowin (approved):

packages/vue-client/src/components/boards/BadukBoardAbstract.vue
JonKo314: Interesting. It is somewhat redundant, but I don't mind having both, a visualization that can handle negative coordinates and coordinates that are all non-negative.

benjaminpjones:
Anything holding this one back? It looks great, and Im hoping we can avoid merge issues by putting it on `main` sooner than later 😋

JonKo314:
Nope, thanks for the reminder.

PR #76, opened 2022-10-01
Let Git handle line endings (LF/CRLF) instead of ESLint

JonKo314:
* Let Git "ensure that text files have their line endings normalized". (see https://www.git-scm.com/docs/gitattributes)
* Tell ESLint to "maintain existing line endings" via Prettier's `endOfLine: "auto"` option. (see https://prettier.io/docs/en/options.html)

Why?
Depending on their machine-wide Git settings Windows users might or might not use CRLF instead of LF when Git checks out files to their filesystem and Git might or might not change those CRLFs to LFs when checking in changes.
* Having ESLint interfere with that can cause lots of errors/warnings or lots of changes to files which Git recognizes as such.
* To avoid having CRLF line endings committed, we can tell Git to normalize the line endings for this repository via the `.gitattributes` file.

Caution:
`* text=auto` makes Git replace CRLF with LF for all files which Git *guesses* to be text files. If Git wrongfully guesses a binary file to be a text file it can damage the file by doing the replacement. If this causes troubles in the future we'd need to e.g. add
`*.ico -text` to `.gitattributes` to tell Git that files with certain file endings are known to be binary files.

### DIFF
--- a/packages/vue-client/.eslintrc.cjs
+++ b/packages/vue-client/.eslintrc.cjs
@@ -16,6 +16,6 @@ module.exports = {
     },
   ],
   rules: {
-    "prettier/prettier": "error",
+    "prettier/prettier": ["error", { endOfLine: "auto" }],
   },
 };
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto


packages/vue-client/.eslintrc.cjs
```diff
@@ -16,6 +16,6 @@ module.exports = {
     },
   ],
   rules: {
-    "prettier/prettier": "error",
+    "prettier/prettier": ["error", { endOfLine: "auto" }],
```
benjaminpjones, line 19: I know you had mentioned adding a config might help VSCode ignore your global config. Does this do that? Or we could move this into its own file if necessary.

benjaminpjones (approved):
Awesome, thanks for doing the research and sharing the links on that!

I'm away from my personal Mac for a few days so I couldn't test there, but I tried this out on Ubuntu and no issues with the linter/VSCode plugins.

Issue #38, opened 2022-07-31
Handle errors and crashes more gracefully


benjaminpjones:
It recently came to my attention that server-side errors aren't handled well (~~try sending a move from the wrong player!~~ fixed that, but there are others). If an async function throws, **the server will crash** and needs to be manually restarted. A couple things we can do:

- [ ] Use a process manager to restart whenever there is a failure.
- [x] catch the unhandled promises
- [ ] ~~Set the right error instead of throwing: (#29)~~

benjaminpjones:
Apparently [Heroku already has an automatic restart policy](https://stackoverflow.com/a/71326832/5001502), so the Process Manager suggestion is moot.  I must have forgotten to put "Heroku" in my search terms before - I was looking at threads like https://stackoverflow.com/a/71326832/5001502.

PR #79, opened 2022-10-10
Eslint server no-floating-promises

benjaminpjones:
This might help with #38, but still a lot of things to do for that task

For those not familiar:

```
async function someAsyncFunct() {
    throw new Error("lol oops");
}

// somewhere else, this will crash the server!!
someAsyncFunc();
```

Now you have to catch these things, which is safer:
```
someAsyncFunc().catch((e) => console.log(`Error: ${e}`));
```


benjaminpjones:
Lol forgot this was hanging out

PR #91, opened 2023-04-01
Select seat on reload

### LINKED ISSUE #85: If player is sitting at a seat, that seat should be selected when the page is reloaded
It rememebers you are sitting at the seat, but not which player was selected (you can sit at multiple seats, hence the selection after sitting)

JonKo314:
If the user sits at exactly one seat in a game,
select that seat when the game is loaded

closes #85


benjaminpjones:
LGTM!

PR #92, opened 2023-04-03
added forms for variant game options panes, fixed size bug in polygonal board helper

merowin:


benjaminpjones:
I haven't read the code yet, but I just tried it out, and looks great!

Regarding "All checks have failed": `yarn lint` should automatically fix most, but not all the errors.  Some (such as [no-unused-vars](https://eslint.org/docs/latest/rules/no-unused-vars)) need to be fixed manually.  Fixing the eslint/prettier errors will get the CI green again.

packages/shared/src/badukWithAbstractBoard/abstractBoard/PolygonalBoardHelper.ts
```diff
@@ -316,11 +316,13 @@ class PolygonalTile {
 }
 
 export function CreatePolygonalBoard(size: number) : Intersection[] {
+    
     const StartTile: PolygonalTile = new PolygonalTile(new Vector2D(0, 0));
     const Tiles: PolygonalTile[] = [StartTile];
     let tileQueue: PolygonalTile[] = [StartTile];
 
-    for (let i = 1; i < (size + 1) / 2; i++) {
+    for (let i = 1; i <= (size - 1) / 2; i++) {
```
benjaminpjones, line 324: Is this a breaking change? Will this affect already-created polygonal boards?

packages/vue-client/src/components/GameCreation/GameCreationForm.vue
```diff
@@ -0,0 +1,87 @@
...
+    <div>
+      <label>Variant:</label>
+      <select v-model="variant">
+        <option v-for="variant in variants" :key="variant">
+          {{ variant }}
+        </option>
+      </select>
+    </div>
+    <label>Config: </label>
+    <template v-if="variant==='baduk' || variant==='phantom' || variant=='capture'">
```
benjaminpjones, line 62: Just a thought..

It might nice if we could have a mapping from Variants to ConfigViews, kind of like how we map variants to boards ([board_map.ts](https://github.com/benjaminpjones/govariants/blob/main/packages/vue-client/src/board_map.ts)).  The benefits would be brevity and separation of concerns.

But that is not a blocker, I think it looks pretty good as-is!

benjaminpjones (approved):

packages/shared/src/badukWithAbstractBoard/abstractBoard/PolygonalBoardHelper.ts
merowin: Good point, yes I believe this would potentially break existing games with polygonal boards.

The "bug" this change tries to fix is not really a problem actually. I just think that the size of the polygonal board grows more naturally with this.

benjaminpjones: Okay good to know, breaking changes are okay - we don't have users to be bothered, just wanted to check!

packages/vue-client/src/components/GameCreation/GameCreationForm.vue
merowin: ok I added a map from variants to the form components, named config_form_map

also ran yarn lint and removed sign restriction from komi inputs.

packages/vue-client/src/assets/main.css
```diff
@@ -33,3 +33,8 @@ a,
     padding: 0 2rem;
   }
 }
+
+.column {
+  display: flex;
+  flex-direction: column;
+}
```
JonKo314, line 40: Somehow I don't like the name of this class. A column can also appear in a grid layout or table layout and those will probably appear in the `main.css` too. We can leave it like this for now, but maybe I should look up some good practices for naming CSS classes.

packages/vue-client/src/components/GameCreation/BadukConfigForm.vue
```diff
@@ -0,0 +1,43 @@
...
+    <label>Komi</label>
+    <input
+      type="number"
+      step="0.5"
+      v-model="config.komi"
+      @change="emitConfigChange"
+      style="width: fit-content"
+    />
+  </div>
+</template>
```
JonKo314, line 43: Why not just add a CSS rule `input { width: fit-content }` inside a `<style>` block for the component?  Don't repeat yourself (DRY).

Or `.awesomeClassName input { width: fit-content }` somewhere for all config components?

packages/vue-client/src/components/GameCreation/GameCreationForm.vue
```diff
@@ -0,0 +1,74 @@
+<script setup lang="ts">
```
JonKo314, line 1: Dear Git, why don't you recognize this file isn't new, but was just moved and modified?
See [How to make git mark a deleted and a new file as a file move?](https://stackoverflow.com/a/433142) if you want to avoid this in the future.

packages/vue-client/src/components/GameCreation/BadukConfigForm.vue
```diff
@@ -0,0 +1,43 @@
...
+  (e: "configChanged", config: BadukConfig): void;
+}>();
+
+function emitConfigChange() {
+  emit("configChanged", config);
+}
+</script>
+
+<template>
+  <div class="column">
```
JonKo314, line 17: Maybe this `<div>` could be a `<form>` with a `@change`, then it wouldn't be necessary to add `@change` to each `<input>`.

JonKo314 (approved):
Thanks for adding this :+1:.

I've left a few notes but nothing important, feel free to ignore them.

packages/vue-client/src/components/GameCreation/GameCreationForm.vue
benjaminpjones: Haha the diff is pretty significant, may have been unavoidable 😅

packages/vue-client/src/assets/main.css
merowin: renamed to config-form-column

packages/vue-client/src/components/GameCreation/BadukConfigForm.vue
merowin: added style block and rule

packages/vue-client/src/components/GameCreation/BadukConfigForm.vue
merowin: done

PR #96, opened 2023-04-30
Time control

merowin:


packages/server/src/time-control.ts
```diff
@@ -0,0 +1,90 @@
...
+
+            case TimeControlType.Invalid:
+                console.error("game with invalid time control type");
+                return;
+        }
+    }
+}
+
+class TimeHandlerParallelMoves implements ITimeHandler {
+    constructor() {}
```
benjaminpjones, line 75: Nit: no need to define empty constructor

packages/server/src/time-control.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export function ValidateTimeControlBase(time_control: unknown): time_control is ITimeControlBase {
+    return (time_control && typeof time_control === "object" && 
+    'moveTimestamps' in time_control && 'remainingMilliseconds' in time_control && 'onThePlaySince' in time_control);
+}
+
+// validation of the config should happen before this is called
+export interface ITimeHandler {
+    handleMove(game: GameResponse, playerNr: number, move: string): Promise<void>;
```
benjaminpjones, line 21: I'll be interested to see the variant where different moves affect the time control differently 😄

benjaminpjones (approved):

packages/server/src/time-control.ts
```diff
@@ -0,0 +1,90 @@
...
+                    console.error(`game with id ${game.id} is missing player nr ${playerNr} in time control`);
+                    return;
+                }
+
+                if (!game.players) {
+                    console.log(`game with id ${game.id} has no players array`);
+                    return;
+                }
+
+                const nextPlayerNr = (playerNr + 1) % (game.players.length + 1);
```
benjaminpjones, line 52: This code will be more reusable if we keep the `nextPlayerNr` logic outside (after all, the containing function `playMove` already has access to [`game_obj.nextToPlay()`](https://github.com/benjaminpjones/govariants/blob/5eef3d775b26fdbba73706cecd2f273f72cd7161/packages/shared/src/abstract_game.ts#L31).

The reason I bring this up is that variants such as Thue-Morse can probably work very similarly to the "sequential" variants.

packages/server/src/time-control.ts
merowin: Right, this is unnecessary at the moment. But I think that it does not hurt to include this - who knows what variants we'll come up with in the future :D

packages/server/src/time-control.ts
merowin: added game object to handle move method and use game_obj.nextToPlay()

packages/server/src/time-control.ts
merowin: removed

JonKo314 (approved):
Looking good, thanks!

PR #108, opened 2023-06-28
dedupe getOnlyMove

benjaminpjones:
Long needed code health change - we currently have 5 copies of this function.

Doesn't really address the data format issue.  It's now apparent that moves will never really be more than one move even for parallel variants.  I'm wondering if we should just store moves as the output of this function, e.g. `{ player: 1; move: "jk"}`. It's a little less compact, but easy to work with and readable.

### DIFF
--- a/packages/server/src/games.ts
+++ b/packages/server/src/games.ts
@@ -3,6 +3,7 @@ import {
   makeGameObject,
   MovesType,
   User,
+  getOnlyMove,
 } from "@ogfcommunity/variants-shared";
 import { ObjectId, WithId, Document } from "mongodb";
 import { getDb } from "./db";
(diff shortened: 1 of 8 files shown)


JonKo314:
> It's now apparent that moves will never really be more than one move even for parallel variants.
Were we thinking a *move* was equal to a *round*? And that all player-moves per round should be gathered in one *move* object? Seems strange now. :smile: 

> I'm wondering if we should just store moves as the output of this function, e.g. { player: 1; move: "jk"}. It's a little less compact, but easy to work with and readable.
Are you talking about disk space usage when you say "less compact"? Or just in general?

benjaminpjones:
> Were we thinking a _move_ was equal to a _round_? And that all player-moves per round should be gathered in one _move_ object? Seems strange now. 😆

I think exactly this. >.< In my head, parallel variants would have been passing around objects like `{1: "aa", 2: "bb", 3: "cc"}`, but it just never happens because players have to submit their moves separately anyway.

> Are you talking about disk space usage when you say "less compact"? Or just in general?

I guess in general.  Disk usage is one, but network usage and test code verbosity increase as well.  I'm not thinking it's something we need to worry about, but it's more than 2x the size of the current structure.

```
{"1":"jk"} (10 bytes)
[1,"jk"] (8 bytes)
{"player":1,"move":"jk"} (24 bytes)
```

If size *is* a concern, the second option above is actually nice.  Smaller and easier to access the elements than `{ 1: "jk" }`

benjaminpjones:
Merged, but a change of data format is probably the right move long term - will work out a migration plan

PR #120, opened 2023-07-01
Create group_utils.ts

### LINKED ISSUE #114: Group utility functions
Need some easy way of getting all the points in a group and related functions.  To start, maybe these:

- flood do (for each stone in group, perform some callback)
- flood fill
- get group
- get border (inner, outer)

Requirements:
 - reasonably efficient
 - Should operate on `Grid` in a generic way.  Maybe a "Fillable" or "GraphLike" interface:
 
```
interface Fillable<IndexT, ValueT> {
    neighbors: (idx: IndexT) => IndexT[];
    at: (idx: IndexT) => ValueT;
    set: (idx: IndexT, value: ValueT) => ValueT;
       ...
}
```

I think #90 has something along these lines, but I haven't looked closely yet.

benjaminpjones:
Flood fill/grouping is a core algorithm of most go variants. This is a set of functions that can operate on the Grid class in a sort of generic way.

Fixes #114

### DIFF
--- a/packages/shared/src/lib/group_utils.ts
+++ b/packages/shared/src/lib/group_utils.ts
@@ -1,0 +1,71 @@
+// Perhaps we can remove some of these functions if it becomes cumbersome to
+// implement these functions every time.  For example, `forEach` can be replaced
+// by map if we ignore the output
+interface Fillable<K, V> {
+  at(index: K): V | undefined;
+  set(index: K, val: V): void;
+  map<V2>(f: (val: V) => V2): Fillable<K, V2>;
+  neighbors(index: K): K[];
+  isInBounds(index: K): boolean;
+  forEach(f: (value: V, index: K) => void): void;
+}
+
+export function getGroup<K, V>(index: K, g: Fillable<K, V>): K[] {
+  if (!g.isInBounds(index)) {
+    return [];
+  }
+
+  const ret: K[] = [];
+
+  // The type assertion (as V) is regrettable, but I can't think
+  // of a better way to satisfy the strict null checks
+  // At any rate, calling floodDo on an out-of-bounds index should be a no-op
+  const starting_value = g.at(index) as V;
+  const visited = g.map(() => false);
+
+  function helper(index: K) {
+    if (g.at(index) !== starting_value || visited.at(index)) {
+      return;
+    }
+    visited.set(index, true);
+    ret.push(index);
+    g.neighbors(index).forEach(helper);
+  }
+  helper(index);
+
+  return ret;
+}
+
+export function floodFill<K, V>(value: V, index: K, g: Fillable<K, V>) {
+  return getGroup(index, g).forEach((index) => g.set(index, value));
+}
+
+export function getOuterBorder<K, V>(group: K[], graph: Fillable<K, V>) {
+  // true if index is in group
+  const group_on_graph = graph.map(() => false);
+  // mark "expanded" group on graph
+  group.forEach((index) =>
+    graph.neighbors(index).forEach((index) => group_on_graph.set(index, true))
+  );
+  // unmark group
+  group.forEach((index) => group_on_graph.set(index, false));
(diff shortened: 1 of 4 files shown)


packages/shared/src/lib/group_utils.ts
```diff
@@ -0,0 +1,71 @@
+// Perhaps we can remove some of these functions if it becomes cumbersome to
+// implement these functions every time.  For example, `forEach` can be replaced
+// by map if we ignore the output
+interface Fillable<K, V> {
+  at(index: K): V | undefined;
```
JonKo314, line 5: What does undefined mean here? Out of bounds or no value?

packages/shared/src/lib/group_utils.ts
```diff
@@ -0,0 +1,71 @@
...
+  if (!g.isInBounds(index)) {
+    return [];
+  }
+
+  const ret: K[] = [];
+
+  // The type assertion (as V) is regrettable, but I can't think
+  // of a better way to satisfy the strict null checks
+  // At any rate, calling floodDo on an out-of-bounds index should be a no-op
+  const starting_value = g.at(index) as V;
```
JonKo314, line 23: If `at()` returned `undefined` for out-of-bounds indices, we could use something like this:
```
const startingValue = g.at(index);
if (startingValue === undefined) {
  return [];
}
```

packages/shared/src/lib/group_utils.ts
benjaminpjones: In the case of `Grid` (the only class that implements `Fillable` at the moment, `at` can return `undefined` for a number of reasons:

```
  const g1 = new Grid(2, 2);
  // undefined, empty slot
  g1.at({ x: 1, y: 1 });

  const g2 = Grid.from2DArray([
    [1, 2],
    [3, 4],
  ]);
  // undefined, out-of-bounds
  g2.at({ x: 0, y: 5 });

  const g3 = Grid.from2DArray([
    [undefined, undefined],
    [undefined, undefined],
  ]);
  // undefined, T is undefined
  g3.at({ x: 0, y: 0 });
```

FWIW, this matches the behavior for native `Array`s (but I'm not opposed to changing it if it makes the developer experience better)

```
const a1 = new Array(5);
a1[0];  // undefined

const a2 = [1, 2, 3];
a2[5];  // undefined

const a3 = [undefined, undefined, undefined];
a3[0];  // undefined
```

packages/shared/src/lib/group_utils.ts
benjaminpjones: Related to my other comment, I would expect to get a non-empty group for in-bounds `undefined`:

```
const my_grid = Grid.from2DArray([
    [undefined, 0, 1],
    [undefined, 0, 1],
    [undefined, 0, 1],
])
getGroup({ x: 0, y: 0 }, my_grid).length  // 3
```

packages/shared/src/lib/group_utils.ts
JonKo314: Thanks!

packages/shared/src/lib/group_utils.ts
JonKo314: Okay, so we'd have to prevent `undefined` to be valid for `V`, I didn't think about that before. There's always `null` to use, but just to get rid of `as V` isn't a great reason to do so.

benjaminpjones: Yep I think that sounds right, and thanks for thinking on solutions for this too.  The only other "solution" I've thought about was to change the typing to say `.at()` returns `T` instead of `T | undefined`.  It would, of course, be a lie, but [it is one of the lies TypeScript makes for array accesses in the name of dev experience](https://stackoverflow.com/questions/50647399/typescript-accessing-an-array-element-does-not-account-for-the-possibility-of-u) 😆

PR #124, opened 2023-07-02
Time control

merowin:


packages/vue-client/src/components/Timer.vue
```diff
@@ -0,0 +1,39 @@
...
+  let minutes = Math.floor(msCopy / (1000 * 60));
+  msCopy = msCopy - minutes * (1000 * 60);
+  let seconds = Math.floor(msCopy / 1000);
+
+  return `${days.toString()}d : ${hours.toString()}h : ${minutes.toString()}m : ${seconds.toString()}s`;
+}
+</script>
+
+<template>
+    {{ msToTime(t) }}
```
JonKo314, line 38: I think [computed properties](https://vuejs.org/guide/essentials/computed.html) are meant for things like this, but not necessary.

packages/vue-client/src/components/Timer.vue
```diff
@@ -0,0 +1,39 @@
...
+function msToTime(ms: number): string {
+  //ToDo: improve, this is kinda horrible
+  let msCopy = ms;
+  let days = Math.floor((msCopy / (1000 * 60 * 60 * 24)))
+  msCopy = msCopy - days * (1000 * 60 * 60 * 24);
+  let hours = Math.floor((msCopy / (1000 * 60 * 60)))
+  msCopy = msCopy - hours * (1000 * 60 * 60);
+  let minutes = Math.floor(msCopy / (1000 * 60));
+  msCopy = msCopy - minutes * (1000 * 60);
+  let seconds = Math.floor(msCopy / 1000);
```
JonKo314, line 31: Maybe it would be less verbose to go the other way round? Calculate seconds first?

merowin: right, I changed it accordingly. I was mainly thinking there is probably a "proper" way of doing this, like a method of Date or so, that would create a formatted string. But for now I'm actually happy with what I got now.

packages/vue-client/src/components/Timer.vue
merowin: I don't know when the lambda function of a computed variable is being called. I've now changed the time to a ref, and added a watch on it.

packages/vue-client/src/components/GameTimer.vue
```diff
@@ -0,0 +1,93 @@
...
+watch(time, (t: number) => {
+  if (t === null) {
+    formattedTime.value = "";
+  } else {
+    formattedTime.value = msToTime(t);
+  }
+});
+
+onMounted(() => resetTimer());
+watch(props, () => resetTimer());
```
benjaminpjones, line 28: I couple things about `watch`:

- watch has an `immediate` option that causes the function to run when the component is created, so that might be able to replace the `onMounted` call. ([docs](https://vuejs.org/guide/essentials/watchers.html#eager-watchers))
- The third argument of the callback is a `cleanup` function, so we could consolidate the `clearInterval` logic and remove the `onUnmounted` call. ([docs](https://vuejs.org/api/reactivity-core.html#watch))

```
watch(props, (value, oldValue, onCleanup) => {
      ...
    const timerIndex = setInterval(...);
      ...
    onCleanup(() => clearInterval(timerIndex));
}, {immediate: true})
```

benjaminpjones (approved):
Awesome stuff - time is hard, but you've structured this quite nicely

packages/vue-client/src/components/GameCreation/TimeControlConfigForm.vue
```diff
@@ -0,0 +1,54 @@
...
+<template>
+  <form @change="emitConfigChange" class="config-form-column">
+    <label>Time Control</label>
+    <select v-model="typeRef">
+      <option :value="null">Unlimited Time</option>
+      <option :value="TimeControlType.Absolute">Absolute</option>
+      <!--<option :value="TimeControlType.Fischer">Fischer</option>   -->
+    </select>
+    <template v-if="typeRef !== null">
+      <label>Main Time</label>
```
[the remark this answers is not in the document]
merowin: Right. Now that I think about it, should I change the unit in the input form to seconds? I doubt anybody wants to specify Milliseconds :sweat_smile:

packages/vue-client/src/components/GameTimer.vue
merowin: Thank you for your insights! I've changed it accordingly.

packages/vue-client/src/components/GameCreation/TimeControlConfigForm.vue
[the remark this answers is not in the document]
merowin: changed the unit of this input to seconds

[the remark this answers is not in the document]
benjaminpjones: Seconds is much better!

PR #127, opened 2023-07-06
upgrade everything

benjaminpjones:
idk if we want a more elegant upgrade strategy, but I figure we're due after one year.  I tested the usual scripts (`yarn build`, `yarn start`, `yarn launch`), but that's about it.

The biggest thing in here is TypeScript 5.  Looks like Mongo had a major version bump too, but I didn't see any breakages.

Motivation for the upgrade was mainly an attempt to resolve these deprecation errors:

Screenshot 2023-07-06 at 5 03 56 PM


packages/vue-client/src/components/boards/ParallelGoBoard.vue
```diff
@@ -109,6 +109,6 @@ const board = computed(() => {
   <MulticolorGridBoard
     :board="board"
     :config="props.config"
-    @click="(pos) => positionClicked(pos)"
```
benjaminpjones, line 112: changed to address implicit any error

packages/vue-client/tsconfig.app.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "@vue/tsconfig/tsconfig.web.json",
```
benjaminpjones, line 2: see [vuejs/tsconfig README: Migrating from TypeScript <5.0](https://github.com/vuejs/tsconfig#migrating-from-typescript--50)

packages/vue-client/tsconfig.config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "@vue/tsconfig/tsconfig.node.json",
```
benjaminpjones, line 2: see [vuejs/tsconfig README: Migrating from TypeScript <5.0](https://github.com/vuejs/tsconfig#migrating-from-typescript--50)

JonKo314 (approved):
Well, `yarn test` had no issue, so I think we're good to go. :+1:

PR #149, opened 2023-08-06
Merge Baduk and abstractAlternatingOnGrid

benjaminpjones:
When this class was conceived, there were only two variants (Baduk and TicTacToe).  A dozen variants later, a couple observations can be made:

- Most variants that might inherit from `abstractAlternatingOnGrid` benefit more from inheriting from `Baduk`.  This is because most variants require the "capture" game mechanic anyway.
- Having the `Baduk` functionality split across classes can make it difficult to find the right entry points when modifying `Baduk`
- We can use composition-based abstractions such as `Grid` and `group_utils` to remove some complex code from `Baduk`

While I remove the abstract class in this PR, I keep the division of the `playMove` method into submethods (`postValidateMove`, `playMoveInternal`, etc) as those have provided useful entry points for subvariants to modify a part of `Baduk`.

---

Before

Screenshot 2023-07-29 at 8 13 11 PM

After

Screenshot 2023-07-29 at 8 12 32 PM


JonKo314:
>     * Having the `Baduk` functionality split across classes can make it difficult to find the right entry points when modifying `Baduk`

I'll trust your experience here, so the restructuring is fine with me. :+1:

PR #154, opened 2023-12-31
Allow player to resign if it's not their turn

benjaminpjones:
Fix bug where players cannot resign if it's not their turn.


benjaminpjones:
Going to close this for now since it could conflict with #152 - I can come back to it later

Issue #155, opened 2024-01-02
Selecting variant after selecting a time control sets it to Unlimited without an indication in the UI


benjaminpjones:
1) Select **Absolute** time control

<img width="322" alt="Screenshot 2024-01-01 at 10 42 44 PM" src="https://github.com/govariantsteam/govariants/assets/25233703/64591328-cdd7-4f80-a7ac-f183aa3cc345">

2) Select another variant

<img width="323" alt="Screenshot 2024-01-01 at 10 43 07 PM" src="https://github.com/govariantsteam/govariants/assets/25233703/8be19540-51ca-43de-bfc3-db7909e6219e">

3) Observe: the time control still says **Absolute**

4) Click Create Game

<img width="231" alt="Screenshot 2024-01-01 at 10 44 01 PM" src="https://github.com/govariantsteam/govariants/assets/25233703/bd4d622f-3547-4b15-b096-9cdaaa2407f9">

5) Observe: there is no game clock (and no time control in the config)

PR #161, opened 2024-01-04
bugfix time control config form

merowin:
this should fix https://github.com/govariantsteam/govariants/issues/155

### DIFF
--- a/packages/vue-client/src/components/GameCreation/GameCreationForm.vue
+++ b/packages/vue-client/src/components/GameCreation/GameCreationForm.vue
@@ -41,13 +41,9 @@ const createGame = async () => {
   }
   //ToDo: add form validation
 
-  if (timeControlConfig !== null) {
-    config.time_control = timeControlConfig;
-  }
-
   const game = await requests.post("/games", {
     variant: variant.value,
-    config: config,
+    config: { ...config, time_control: timeControlConfig },
   });
   router.push({ name: "game", params: { gameId: game.id } });
 };
--- a/packages/vue-client/src/components/GameCreation/GameCreationForm.vue
+++ b/packages/vue-client/src/components/GameCreation/GameCreationForm.vue
@@ -17,7 +17,7 @@ interface IConfigWithOptionalTimeControl {
 const variants: string[] = getVariantList();
 const variant: Ref<string> = ref(variants.length ? variants[0] : "");
 let config: IConfigWithOptionalTimeControl;
-const timeControlConfig: ITimeControlConfig | null = null;
+let timeControlConfig: ITimeControlConfig | null = null;
 const configString: Ref<string> = ref("");
 const variantConfigForm = computed(() => config_form_map[variant.value]);
 watch(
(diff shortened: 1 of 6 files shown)


packages/vue-client/src/components/GameCreation/GameCreationForm.vue
```diff
@@ -33,14 +33,18 @@ const createGame = async () => {
...
+      typeof timeControlConfig.mainTimeMS !== "number")
   ) {
     console.log("game creation form is invalid");
     return;
   }
   //ToDo: add form validation
 
+  if (timeControlConfig !== null) {
+    config.time_control = timeControlConfig;
+  }
```
benjaminpjones, line 46: Nit: I don't love mutating `config` - if for some reason we stayed on the page instead of redirecting (maybe the `POST` fails?).  The time control will remain on `config`.

Then, if we set the config to "Unlimited" again, the wrong time control will be used.

---

An alternative is to leave config unmutated, but pass `{ ...config, time_control: timeControlConfig }` to the `POST` call (`...` is the [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#spread_in_object_literals))

benjaminpjones (approved):

packages/vue-client/src/components/GameCreation/GameCreationForm.vue
merowin: wait, why was that nonsense? I think your suggestion sounded good:

pass { ...config, time_control: timeControlConfig } to the POST call

benjaminpjones: Hmm I guess immutability is still not bad idea, but I saw the other solution, and didn't want to be an FP zealot 😅 

Also there is a slight difference between

```
{ ...config, time_control: timeControlConfig }
```

and 

```
delete config["time_control"]
```

because the former will send a config like:

```
{ width: 19, height: 19, time_control: null }
```

And I'm not sure if that matters.

merowin: I see, but you make a good point. I've added your suggestion together with a null check, and moved the type check function to shared (because it was really used on both ends).

In retrospective, I think I should have avoided working so much with undefined when I started, just setting it to null is simpler as opposed to not setting the property at all.

I'm learning ^^

merowin: > didn't want to be an FP zealot 😅

by the way, I have no idea what that means xD

JonKo314: I'd guess FP means Functional Programming, but I might be totally wrong.

JonKo314 (approved):

packages/vue-client/src/components/GameCreation/GameCreationForm.vue
benjaminpjones: Yep, [functional programming](https://en.wikipedia.org/wiki/Functional_programming)!  (all variables are immutable in FP)

PR #164, opened 2024-01-06
prototype for drift go

merowin:
This is a prototype for a variant idea:

_Drift Go
After every move, the board shifts in a toroidal-like translation, e.g. the top row becomes the bottom row._

In this implementation, groups that lose their last liberty as a result of the shift are all removed simultaneously and grant no captures.

This is probably not the most efficient implementation (the shift can be represented by only two numbers without re-assigning the whole board), but after testing it, the performance is not bad.


benjaminpjones:
This is sweet!

PR #165, opened 2024-01-07
Support empty config for Pyramid Go

### LINKED ISSUE #163: Game creation form crashes when selecting pyramid variant
Not sure what exactly causes the error.

Steps to reproduce:
Select pyramid in the variant dropdown in the game creation form -> the form collapses and console shows error "TypeError: t is undefined"

benjaminpjones:
Fixes #163 

We had discussed some refactor of the defaut configs/game map, but actually, here is a 4 line fix, and I've also added a test which should prevent this class of bugs in the future.

(I'm still supportive of a refactor, but that takes more care and I wanted to get this variant working again)


merowin:
nice!

merowin (approved):

PR #166, opened 2024-01-07
Parallel Board: Don't stage passes at 'pa'

### LINKED ISSUE #162: in parallel variant, pass is displayed as a move on "pa" (sgf representation)
grafik

Steps to reproduce:

Create 19x19 parallel game and pass. The player who passed sees a stone placed at "pa", and next round all players see a "last move indicator" at "pa".

benjaminpjones:
Fixes #162

### DIFF
--- a/packages/vue-client/src/components/boards/ParallelGoBoard.vue
+++ b/packages/vue-client/src/components/boards/ParallelGoBoard.vue
@@ -10,7 +10,13 @@ function forEachMoveModifyStone(
 ) {
   Object.keys(moves).forEach((player_str) => {
     const player = Number(player_str);
-    const move = Coordinate.fromSgfRepr(moves[player]);
+    const move_str = moves[player];
+    if (move_str.length !== 2) {
+      // special moves like "pass" and "resign"
+      return;
+    }
+    const move = Coordinate.fromSgfRepr(move_str);
+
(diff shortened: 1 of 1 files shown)


packages/vue-client/src/components/boards/ParallelGoBoard.vue
```diff
@@ -10,7 +10,13 @@ function forEachMoveModifyStone(
 ) {
   Object.keys(moves).forEach((player_str) => {
     const player = Number(player_str);
-    const move = Coordinate.fromSgfRepr(moves[player]);
+    const move_str = moves[player];
+    if (move_str.length !== 2) {
+      // special moves like "pass" and "resign"
+      return;
+    }
+    const move = Coordinate.fromSgfRepr(move_str);
```
benjaminpjones, line 18: This makes me wonder if this is actually a bug in `fromSgfRepr`.

Current behavior:

- output { x: 15, y: 0 }

Potentially improved behavior:

- output {x: -1, y: -1}
- Throw an error ("pass" is not how passes are represented in SGF anyway)

merowin (approved):
Looks good. I think this function (fromSgfRepr) is called very often, so its desirable to keep it as light-weight as possible. We'll just have to keep in mind to filter out the special moves first, like you did here.

PR #169, opened 2024-01-12
add prettier config, lint project

merowin:
added a prettier config, then run yarn lint in the root directory.

I'm hoping that this will replace all newline stuff (CRLF -> LF), and maybe there will be less changes on lint in future. But I don't know much about this, so ...

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,5 +19,9 @@
     "packages": [
       "packages/*"
     ]
+  },
+  "prettier": {
+    "trailingComma": "all",
+    "endOfLine": "lf"
   }
 }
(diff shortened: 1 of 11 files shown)


JonKo314 (approved):
`"trailingcomma": "all"` and `"endOfLine": "lf"` should be enabled by default, so I don't know why specifying it explicitly helps, but I'm glad if it does.

If you still run into troubles with LF/CRLF, maybe check your Git settings if you haven't done that yet: https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

benjaminpjones:
`"trailingcomma": "all"` was introduced fairly recently (see [Prettier 3.0 Breaking Changes](https://prettier.io/blog/2023/07/05/3.0.0.html#change-the-default-value-for-trailingcomma-to-all-11479httpsgithubcomprettierprettierpull11479-by-fiskerhttpsgithubcomfisker-13143httpsgithubcomprettierprettierpull13143-by-sosukesuzukihttpsgithubcomsosukesuzuki))

Which explains why vue-client was using the old style. Might also be worth an upgrade to 3 for consistency.

JonKo314:
Oh, I see, thanks!

I wouldn't mind updating to v3.

PR #171, opened 2024-01-16
Games filter (prototype)

merowin:
Issue: https://github.com/govariantsteam/govariants/issues/172

Prototype for a way to filter for games of one variant, and / or games where the user has taken a seat.

grafik

### DIFF
--- a/packages/server/src/games.ts
+++ b/packages/server/src/games.ts
@@ -6,7 +6,7 @@ import {
   getOnlyMove,
   HasTimeControlConfig,
 } from "@ogfcommunity/variants-shared";
-import { ObjectId, WithId, Document } from "mongodb";
+import { ObjectId, WithId, Document, Filter } from "mongodb";
 import { getDb } from "./db";
 import { io } from "./socket_io";
 import { getTimeoutService } from "./index";
@@ -31,7 +31,7 @@ export async function getGames(
   user_id: string | null,
   variant: string | null,
 ): Promise<GameResponse[]> {
-  let dbFilter: Record<string, any> = {};
+  const dbFilter: Filter<Document> = {};
   if (user_id) {
     dbFilter["players.id"] = user_id;
   }
--- a/packages/server/src/games.ts
+++ b/packages/server/src/games.ts
@@ -23,13 +23,15 @@ export function gamesCollection() {
 /**
  * @param count number of games to return (default = 10, max = 100)
  * @param offset number of games to skip (default = 0)
+ * @param user_id filter for games where this user has taken a seat
  */
 export async function getGames(
   count: number,
   offset: number,
+  user_id: string | undefined,
 ): Promise<GameResponse[]> {
   const games = gamesCollection()
-    .find()
+    .find(!user_id ? {} : { "players.id": user_id })
     .sort({ _id: -1 })
     .skip(offset || 0)
     .limit(Math.min(count || 10, 100))
@@ -88,7 +90,7 @@ export async function playMove(
   game_id: string,
   moves: MovesType,
   user_id: string,
-) {
+): Promise<GameResponse> {
   const game = await getGame(game_id);
 
   // Verify that moves are legal
--- a/packages/server/src/games.ts
+++ b/packages/server/src/games.ts
@@ -28,10 +28,19 @@ export function gamesCollection() {
 export async function getGames(
   count: number,
   offset: number,
-  user_id: string | undefined,
+  user_id: string | null,
+  variant: string | null,
 ): Promise<GameResponse[]> {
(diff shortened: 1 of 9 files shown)


merowin:
relevant documentation: https://www.mongodb.com/docs/manual/tutorial/query-arrays/

merowin:
Added filter for variant

packages/shared/src/api_types.ts
```diff
@@ -19,3 +19,8 @@ export interface UserResponse {
   login_type: "guest" | "persistent";
   username?: string;
 }
+
+export type GamesFilter = {
+  user_id: string | null;
+  variant: string | null;
```
benjaminpjones, line 25: Nit: any reason not to make these optional?

```suggestion
  user_id?: string;
  variant?: string;
```

I think that would make it easier to represent empty/incomplete filters:

```
const filter: Ref<GamesFilter> = ref({ user_id: null, variant: "" });
```

vs.

```
const filter: Ref<GamesFilter> = ref({});
```

benjaminpjones (approved):
Looks good! This will be very nice to have.

A couple nits on the filter object/API since I can see the number of filter options growing over time.

Additionally, do you have any thoughts on indexes now that we are doing operations that depend on more than just ID?  Or we could just wait and see.. our database is pretty small right now 😅

packages/vue-client/src/assets/main.css
```diff
@@ -44,3 +44,8 @@ a,
   display: flex;
   flex-direction: column;
 }
+
+.labelledContainer {
+  display: flex;
+  flex-direction: column;
+}
```
JonKo314, line 51: This seems odd to me. What does a container being *labelled* have to do with it having a flex layout applied?

packages/vue-client/src/components/GamesFilterForm.vue
```diff
@@ -0,0 +1,61 @@
...
+} from "@ogfcommunity/variants-shared";
+import { type Ref, ref, computed } from "vue";
+
+const emit = defineEmits<{ (e: "filterChange", filter: GamesFilter): void }>();
+function emitFilter() {
+  emit("filterChange", filter.value);
+}
+
+const variants: string[] = getVariantList();
+const selectedVariant: Ref<string | ""> = ref("");
```
JonKo314, line 15: Why `string | ""`? Isn't `""` a `string` anyway?

JonKo314 (approved):

packages/shared/src/api_types.ts
merowin: Yeah that looks cleaner. The reason is that I learned there is a subtle difference of interpretation of undefined and null, that I'm thinking of when deciding between using undefined / null (e.g. for something optional).

null <-> purposefully set to "no value"
undefined <-> not set (yet)

see e.g. [https://www.geeksforgeeks.org/undefined-vs-null-in-javascript/](url)

But I don't feel strongly about this, so I'll change it as you suggested.

packages/vue-client/src/assets/main.css
merowin: Mainly that I believe this is useful for building the UI. Whenever there's a label to an input, we can wrap them in such a div with this class to display the label on top.

packages/vue-client/src/components/GamesFilterForm.vue
merowin: Yeah 😅 ... I used to have string | null here, consistently with the type in the GamesFilter object. But somehow I failed at making an <option> with value = null, so I decided to use "" instead. I left it in the type explicitly like this, to emphasize that "" has a special meaning, but I'm unhappy about the whole construction here.

packages/server/src/games.ts
```diff
@@ -23,13 +23,24 @@ export function gamesCollection() {
 /**
  * @param count number of games to return (default = 10, max = 100)
  * @param offset number of games to skip (default = 0)
+ * @param user_id filter for games where this user has taken a seat
  */
 export async function getGames(
   count: number,
   offset: number,
+  user_id: string | null,
```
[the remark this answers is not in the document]
merowin: done, thanks!

packages/shared/src/api_types.ts
merowin: made optional

packages/vue-client/src/assets/main.css
merowin: You are right though, it can have a generally applicable name -> renamed to col
Thanks!

packages/vue-client/src/components/GamesFilterForm.vue
JonKo314: I see, looks like `null` is indeed not possible.

PR #175, opened 2024-01-26
Bugfix Polygonal board

merowin:
This aims to fix a bug of polygonal board where tile connections are missing for size >= 7. As a result overlapping tiles were created, significantly slowing down the rendering of the board (e.g. in the games list, very annoying).

Previously board size 20:
grafik

after fix:
bugfix_1

Note how the overlapping tiles result in thicker lines.

I also added a test case for this. This change likely breaks old games, but only those affected by the bug (I believe).

As a general note, the whole polygonal board helper could probably be simplified a lot, using this idea with cyclic direction parameter.

### DIFF
--- a/packages/shared/src/variants/__tests__/polygonal.test.ts
+++ b/packages/shared/src/variants/__tests__/polygonal.test.ts
@@ -1,0 +1,11 @@
+import { CreatePolygonalBoard } from "../badukWithAbstractBoard/abstractBoard/PolygonalBoardHelper";
+
+test("Polygonal board has correct number of tiles", () => {
+  const numberOfFullRings = 20;
+
+  const board = CreatePolygonalBoard(2 * numberOfFullRings - 1);
+  const expectedTiles =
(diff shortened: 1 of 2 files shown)


JonKo314 (approved):
In general there seems to be some room for refactoring in `PolygonalBoardHelper.ts`. For example I think the `ConnectTowards` method could be written like this:
```
ConnectTowards(tile: PolygonalTile, direction: 0 | 1 | 2 | 3 | 4 | 5) {
  [
    this.ConnectToRight, this.ConnectToTopRight, this.ConnectToTopLeft,
    this.ConnectToLeft, this.ConnectToBottomLeft, this.ConnectToBottomRight,
  ][direction](tile, false);
}
```

Although the methods `ConnectToRight`, `ConnectToTopRight`, etc. are pretty similar too, so they could probably be replaced by one generalized method.

But refactoring might not be worth it, if we don't have to change that code again. I think we should keep it in mind though.

benjaminpjones (approved):
Love the inclusion of a test with this!

PR #176, opened 2024-01-27
99 anchor the nav bar

merowin:
Here is an attempt to anchor the nav bar (to the top) an rework its style. We made some layout changes at high-level of the dom structure, so I'm looking forward to your review.

- Anchor the nav bar to top and make it sticky
- Replace "Home" with the Logo
- Style the nav entries consistently

## Home:
grafik
## About:
grafik
## Login:
grafik
## Mobile:
grafik

My partner and I worked on this together :-)

### DIFF
--- a/packages/vue-client/src/assets/main.css
+++ b/packages/vue-client/src/assets/main.css
@@ -7,72 +7,6 @@
   height: 100%;
 }
 
-nav {
-  width: 100%;
-  text-align: left;
-  font-size: 1rem;
-  position: sticky;
-  top: 0;
-  z-index: 999;
-  background-color: var(--color-background-soft);
-  display: flex;
-  flex-wrap: wrap;
-  justify-content: space-between;
-  box-shadow: 0px 0px 5px var(--color-shadow);
-  margin-bottom: 5px;
-
-  div {
-    display: flex;
-    align-items: center;
-  }
-
-  .navLogo {
-    width: 2rem;
-    height: 2rem;
-  }
-
-  .navElement {
-    padding: 0.6rem 1.2rem;
-    height: 100%;
-  }
-
-  a {
-    display: flex;
-    align-items: center;
-    text-decoration: none;
-    color: var(--color-text);
-
-    &.router-link-exact-active {
-      background-color: var(--color-primary-transparent);
-    }
-
-    &:hover {
-      transition: 0.4s;
-      background-color: var(--color-primary-transparent);
-    }
-
-    &:visited {
-      color: var(--color-text);
-    }
-  }
-
-  button {
-    border: none;
-    background-color: inherit;
-    font-size: inherit;
-    font-weight: inherit;
-    font-family: inherit;
-    text-wrap: nowrap;
-    cursor: pointer;
-    color: var(--color-text);
-
-    &:hover {
-      transition: 0.4s;
-      background-color: var(--color-warn-transparent);
-    }
-  }
-}
-
 .pageWrapper {
   max-width: 1280px;
   margin: 0 auto;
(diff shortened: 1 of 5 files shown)


benjaminpjones (approved):
Looks beautiful! You and your partner have some nice CSS skills!

packages/vue-client/src/assets/main.css
```diff
@@ -1,43 +1,96 @@
...
   font-weight: normal;
+  display: grid;
+  grid-template-rows: max-content 1fr;
+  height: 100%;
 }
 
-@media (max-width: 540px) {
-  #app {
-    padding: 0.2rem;
+nav {
```
JonKo314, line 10: I think this belongs inside the `<style>` of the `App.vue` component, because that's where the `<nav>` is declared too. Is there any reason to have it in the global `main.css`?

packages/vue-client/src/assets/main.css
```diff
@@ -1,43 +1,96 @@
...
+      background-color: var(--color-warn-transparent);
+    }
   }
 }
 
-a,
-.green {
-  text-decoration: none;
-  color: hsla(160, 100%, 37%, 1);
-  transition: 0.4s;
```
JonKo314, line 21: I can see this has been moved into the `nav {}` block for `<a>` elements, but what about `<a>` elements outside of that (like those in the `AboutView`). Is this rule now gone for those?

packages/vue-client/src/assets/main.css
```diff
@@ -1,43 +1,96 @@
...
 }
 
-@media (max-width: 540px) {
-  #app {
-    padding: 0.2rem;
+nav {
+  width: 100%;
+  text-align: left;
+  font-size: 1rem;
+  position: sticky;
```
JonKo314, line 14: Nice, I didn't know `position: sticky` exists.

JonKo314 (approved):
Looks really cool, thanks to both of you!

packages/vue-client/src/assets/main.css
merowin: We (my partner and I) are unsure how hyperlinks to external sites should be styled. I guess it depends on the user expectations and accessibility (e.g. underlined, color change after visited). So maybe for now we could use the default? Note that there are some resources about that, e.g. https://developer.mozilla.org/en-US/docs/Learn/Accessibility/CSS_and_JavaScript

Nav bar links however don't necessarily need to adhere to any standards, so I believe that it makes sense to separate them as such.

Although if you feel strongly about keeping the old style, I'll revert the change.

packages/vue-client/src/assets/main.css
merowin: Moved the nav style class into App.vue
The style in App.vue is scoped, so we left the combined style selectors that are also needed in UserNav in main.css
Thank you for the suggestion.

packages/vue-client/src/assets/main.css
merowin: to clarify; yes this pr (currently) would change the style of all links (which is outside of the issues scope).
Nav links are styled according to the a.navElement
all other link styles are reverted to default <- But I don't mind changing this to as it was before

JonKo314: I think the default blue and purple don't really fit our current color scheme, underlining the links is probably good though. But no, I don't feel strongly about this :sweat_smile:.

merowin: How about this proposal: We change the links outside the nav bar back to their original color but add an underline. 

Additionally (but optionally) proposal: We change the color of a visited link to a darker / lighter shade of green (depending on the theme).

merowin: Actually, for this pr, I've decided just to revert this change. I still think this is worth thinking about, though, for the future.

merowin: Ok, I'll do that then :-)

We can always revert these changes later, if it turns out no-good. I don't mind deleting my code ;-)

benjaminpjones:
FYI this PR seems to have introduced a layout issue for games with many players.  Details in https://github.com/govariantsteam/govariants/issues/180

PR #177, opened 2024-01-29
set cookie max age to 180 days

### LINKED ISSUE #173: Make login persistent especially on mobile
I've noticed I always get logged out within 1 day on the Brave mobile browser, even when trying to keep a correspondence game open in a tab. So I have to go through a tedious process of logging in and backing up a few pages then reloading to find my game again. I am guessing it's because the site uses the default "session" cookie expiration here:

https://github.com/govariantsteam/govariants/blob/main/packages/server/src/index.ts#L98

And I guess mobile browsers expire "session" cookies at random times. Meanwhile Chrome-based desktop browsers don't expire session cookies by default ever:

https://bugs.chromium.org/p/chromium/issues/detail?id=128513

So it's not like we're getting any extra security from this option.

Can we give the session cookie a definite (and long) lifetime?

merowin:
closes #173

### DIFF
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -98,7 +98,7 @@ app.use(
     cookie: {
       sameSite: "strict",
       secure: "auto", // TODO: See https://www.npmjs.com/package/express-session,
-      maxAge: 86_400_000 * 120, // 120 days
+      maxAge: 86_400_000 * 180, // 180 days
     },
     store: MongoStore.create({ client: getDb() }),
   }),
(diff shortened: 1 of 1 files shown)


JonKo314 (approved):

packages/server/src/index.ts
```diff
@@ -97,7 +97,8 @@ app.use(
     saveUninitialized: false,
     cookie: {
       sameSite: "strict",
-      secure: "auto", // TODO: See https://www.npmjs.com/package/express-session
+      secure: "auto", // TODO: See https://www.npmjs.com/package/express-session,
+      maxAge: 86_400_000 * 180, // 180 days
```
benjaminpjones, line 101: Lol I wonder under what circumstances millisecond precision is necessary...

benjaminpjones (approved):

Issue #160, opened 2024-01-04
Next to play Indicator


merowin:
It would be great if the seats in the game view indicate which players are currently on the play, e.g. by highlighting the seats.

This is especially helpful for Thue-Morse Go, but also multiplayer variants and parallel move variants.

The method nextToPlay of a game object can be used for this.

PR #178, opened 2024-02-01
Next to play indicator

evizam:
merowin and I added a next to play indicator by
- highlighting the seats of the players whose turn it is with a shadow and bolder text
- displaying a list of players to move below or next to the seat list

We also changed the styling of the seat list to make a long list of players wrap to the next column if there's space.

See issue #160 

## Baduk:
grafik
grafik

## Parallel / Fractional:
grafik
grafik

### DIFF
--- a/packages/vue-client/src/components/SeatComponent.vue
+++ b/packages/vue-client/src/components/SeatComponent.vue
@@ -11,12 +11,15 @@ defineProps<{
   player_n: number;
   selected?: number;
   time_control: IPerPlayerTimeControlBase | null;
+  is_players_turn: boolean;
 }>();
 </script>
 
 <template>
   <div
-    v-bind:class="`seat ${selected === player_n ? 'selected' : ''}`"
+    v-bind:class="`seat ${selected === player_n ? 'selected' : ''} ${
+      is_players_turn ? 'to-move' : ''
+    }`"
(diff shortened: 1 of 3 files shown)


packages/vue-client/src/components/SeatComponent.vue
```diff
@@ -11,12 +11,15 @@ defineProps<{
...
+  is_players_turn: boolean;
 }>();
 </script>
 
 <template>
   <div
-    v-bind:class="`seat ${selected === player_n ? 'selected' : ''}`"
+    v-bind:class="`seat ${selected === player_n ? 'selected' : ''} ${
+      is_players_turn ? 'to-move' : ''
+    }`"
```
JonKo314, line 22: This looked a bit verbose to me, so I checked if Vue has a nicer way to do this. And there's this: https://vuejs.org/guide/essentials/class-and-style.html

evizam: Thanks for the suggestion!
I changed it. Is is okay like this?

JonKo314: Yes, that looks nicer, thanks. :slightly_smiling_face:

benjaminpjones:
This is great, thanks!

PR #185, opened 2024-02-08
Time control continued

merowin:
Improve clock UI (with localised units)
Move time handlers into their own file
Make sure the time handler map doesn't cause a crash (for variants that haven't been added to the map yet)
Fix round transition logic (using the new round method)


merowin:
It doesn't show easily on the diff, here's some notes about what I did:

I wanted to make sure that the time handler map doesn't cause a crash for variants that are not added yet.

Previously usage:
`const timeHandler = new timeControlHandlerMap[game.variant]();`

New:
```
const timeHandlerConstructor = timeControlHandlerMap[game.variant];

if (timeHandlerConstructor) {
  const timeHandler = new timeHandlerConstructor();
  ...
}
```

Apparently this causes an error in the pipeline.

I've also added a variable "isRoundTransition" in the handleMove method, to fix the round transition detection. This allows e.g. the parallel handler to correctly work for parallel variants where each round a different set of players is on the play.

packages/server/src/time-control/timeout.ts
```diff
@@ -104,22 +104,25 @@ export class TimeoutService {
   ): void {
     const timeoutResolver = async () => {
       const game = await getGame(gameId);
-
-      const timeoutMove: MovesType = { [playerNr]: "timeout" };
-      game.moves.push(timeoutMove);
       let timeControl = game.time_control;
 
       // this next part is somewhat duplicated from the playMove function
```
benjaminpjones, line 109: I think this is code is looking a lot closer to `playMove()` after this PR.  Is there any reason not to just call `playMove()` from here?

benjaminpjones (approved):
Traveling now, so won't have time to test until later, but code looks good to me.  I like the changes to file structure!

merowin:
> Traveling now, so won't have time to test until later, but code looks good to me. I like the changes to file structure!

I always test my stuff briefly, so please don't feel obligated to test my PRs (I miss stuff sometimes, but usually we can just fix them later)

That being said, I should add test cases for time control huh 😄 

Thank you for the review, I'll take a closer look after work!

JonKo314 (approved):

packages/server/src/time-control/timeout.ts
merowin: I think that would be possible, only I need to circumvent some of the validation.

For example there is a check whether the seat (that is playing a move) is occupied. So when this seat times out, we automatically play a timeout move for it, and it should go through even if the seat is not occupied.

I could add a boolean input to playMove() that disables this validation check (possibly among others).

packages/server/src/time-control/time-handler-sequencial.ts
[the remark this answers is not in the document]
merowin: renamed, thanks for catching that

packages/server/src/time-control/timeout.ts
merowin: I've taken another look, and have a new proposal on how this could work. But I'd like to make a new PR for that change.

PR #186, opened 2024-02-11
Board with trihexagonal tiling

merowin:
This adds a triangular board with trihexagonal tiling.

size 6:
grafik

size 7:
grafik

size 8:
grafik

size 19:
grafik

### DIFF
--- a/packages/shared/src/variants/badukWithAbstractBoard/abstractBoard/TrihexagonalBoardHelper.ts
+++ b/packages/shared/src/variants/badukWithAbstractBoard/abstractBoard/TrihexagonalBoardHelper.ts
@@ -9,8 +9,8 @@ export class TrihexagonalBoardHelper {
   constructor() {
     this.x_base_vector = new Vector2D(1, 0);
     this.y_base_vector = new Vector2D(
-      Math.sin(Math.PI / 6),
-      Math.cos(Math.PI / 6),
+      Math.cos(Math.PI / 3),
+      Math.sin(Math.PI / 3),
     );
   }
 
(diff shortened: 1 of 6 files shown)


merowin:
The grid utils are very nice! 😄

packages/shared/src/variants/badukWithAbstractBoard/abstractBoard/TrihexagonalBoardHelper.ts
```diff
@@ -0,0 +1,76 @@
...
+
+        return intersection;
+      })
+      .to2DArray()
+      .flat()
+      .filter((value) => value !== null)
+      .map((intersection, index) => {
+        intersection!.Identifier = index;
+        return intersection!;
+      });
```
benjaminpjones, line 74: Very minor nit: You can make the filter callback a [type guard](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) (`value is Intersection`), eliminating the need for the non-null assertion (`!`)

```suggestion
      .filter((value): value is Intersection => value !== null)
      .map((intersection, index) => {
        intersection.Identifier = index;
        return intersection;
      });
```

From [typescript-eslint/no-non-null-assertion](https://typescript-eslint.io/rules/no-non-null-assertion/):

> TypeScript's `!` non-null assertion operator asserts to the type system that an expression is non-nullable, as in not null or undefined. Using assertions to tell the type system new information is often a sign that code is not fully type-safe. It's generally better to structure program logic so that TypeScript understands when values may be nullable.

benjaminpjones (approved):
This looks awesome :)

> The grid utils are very nice! 😄

Ha! I wouldn't have thought of this as a Grid, but glad it proved useful!

packages/shared/src/variants/badukWithAbstractBoard/abstractBoard/TrihexagonalBoardHelper.ts
merowin: Oh nice, I didn't know that could be used like this. I've changed it accordingly, thank you for the hint!

PR #188, opened 2024-02-12
set id of intersections in createRthBoard

merowin:
I love that Fractional (and AbstractBaduk) work with arbitrary graphs! For the Rhombitrihexagonal board, it seems that the intersection ids were not set after conversion. This is an attempt to fix it.

### DIFF
--- a/packages/shared/src/lib/abstractBaduk/badukIntersection.ts
+++ b/packages/shared/src/lib/abstractBaduk/badukIntersection.ts
@@ -12,8 +12,8 @@ export class BadukIntersection<
 > extends Intersection {
   stone: TStone | null;
 
-  constructor(v: Vector2D) {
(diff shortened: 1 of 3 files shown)


JonKo314 (approved):
Thanks for spotting and fixing that. I think it shouldn't be too hard to have the constructor require the `id`, that would reduce the risk of forgetting to set it.

merowin:
> Thanks for spotting and fixing that. I think it shouldn't be too hard to have the constructor require the `id`, that would reduce the risk of forgetting to set it.

Right, I changed it accordingly. Before you said so, I wasn't sure whether I should try to do that or not. For the class Intersection, I remember intentionally setting the Identifier later outside of the constructor, although I don't remember the reason ... 😅

PR #196, opened 2024-02-14
refactor play move, fix timer display delay

merowin:
This tries to simplify and reduce duplicated code in the timeout service as we briefly talked about in https://github.com/govariantsteam/govariants/pull/185

My approach is to split the playMove method into two parts: 
- playMove does the user specific validations and then calls ...
- handleMoveAndTime which does the rest (validate the move, update time control, schedule timeouts, update db game, notify clients)

Then I can reuse this second method in the timeout service, where we should skip the user specific validation (if the seat was vacated in the meantime, it still times out).

I've also included another fix, because I noticed this when testing: Over a couple of minutes, the visual timer was running a bit behind. I believe this was due to the fact that I set an interval which decreases the player time by 1s every 1s. However this was delayed by interval callback.
To counteract this, the timer now remembers the timestamp of the last visual update, and reduces the displayed time accordingly. I've tested this, and indeed seems to be much better.

### DIFF
--- a/packages/server/src/games.ts
+++ b/packages/server/src/games.ts
@@ -102,17 +102,6 @@ export async function playMove(
 ): Promise<GameResponse> {
   const game = await getGame(game_id);
 
-  // Verify that moves are legal
-  const game_obj = makeGameObject(game.variant, game.config);
-  game.moves.forEach((moves) => {
-    const { player, move } = getOnlyMove(moves);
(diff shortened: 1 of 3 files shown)


benjaminpjones (approved):
Sorry for the delay in looking at this, changes look good to me!

merowin:
no problem, thank you for reviewing!

Issue #190, opened 2024-02-13
CSS warnings


benjaminpjones:
I recently ran `yarn build` and noticed there are a lot of CSS warnings.  Not sure if we need to fix these, or just update a config to reflect different supported targets.

```
➤ YN0000: ▲ [WARNING] A nested style rule cannot start with "button" because it looks like the start of a declaration [css-syntax-error]
➤ YN0000: 
➤ YN0000:     <stdin>:35:0:
➤ YN0000:       35 │ button {
➤ YN0000:          │ ~~~~~~
➤ YN0000:          ╵ :is(button)
➤ YN0000: 
➤ YN0000:   To start a nested style rule with an identifier, you need to wrap the identifier in ":is(...)" to prevent the rule from being parsed as a declaration.
➤ YN0000: 
➤ YN0000: 
➤ YN0000: ▲ [WARNING] CSS nesting syntax is not supported in the configured target environment ("chrome87", "edge88", "es2020", "firefox78", "safari14") [unsupported-css-nesting]
➤ YN0000: 
➤ YN0000:     <stdin>:54:0:
➤ YN0000:       54 │ .left-container[data-v-8d688e28] {
➤ YN0000:          ╵ ^
➤ YN0000: 
➤ YN0000: 
➤ YN0000: ▲ [WARNING] CSS nesting syntax is not supported in the configured target environment ("chrome87", "edge88", "es2020", "firefox78", "safari14") [unsupported-css-nesting]
➤ YN0000: 
➤ YN0000:     <stdin>:59:0:
➤ YN0000:       59 │ .right-container[data-v-8d688e28] {
➤ YN0000:          ╵ ^
➤ YN0000: 
➤ YN0000: 
➤ YN0000: rendering chunks...
➤ YN0000: warnings when minifying css:
➤ YN0000: ▲ [WARNING] A nested style rule cannot start with "div" because it looks like the start of a declaration [css-syntax-error]
➤ YN0000: 
➤ YN0000:     <stdin>:15:0:
➤ YN0000:       15 │ div[data-v-110f0678] {
➤ YN0000:          │ ~~~
➤ YN0000:          ╵ :is(div)
➤ YN0000: 
➤ YN0000:   To start a nested style rule with an identifier, you need to wrap the identifier in ":is(...)" to prevent the rule from being parsed as a declaration.
➤ YN0000: 
➤ YN0000: 
[...]

PR #197, opened 2024-02-16
wrap nested selectors in :is

merowin:
issue: #190 

This fixes all warnings I'm getting, which is not all unfortunately.

I believe we can also fix the warnings about nested css syntax not being supported on some configured targets, by removing the nested css syntax (i.e.

```
.myContainer {
  &.leftContainer { ... }
}
```
->
`.myContainer.leftContainer { ... }`

That being said, I think that nested css syntax is quite handy in some cases (e.g. for deeply nested components)

### DIFF
--- a/packages/vue-client/src/App.vue
+++ b/packages/vue-client/src/App.vue
@@ -35,7 +35,7 @@ nav {
   box-shadow: 0px 0px 5px var(--color-shadow);
   margin-bottom: 5px;
 
-  div {
+  :is(div) {
     display: flex;
     align-items: center;
   }
--- a/packages/vue-client/src/components/GamesFilterForm.vue
+++ b/packages/vue-client/src/components/GamesFilterForm.vue
@@ -51,10 +51,10 @@ const filter = computed(() => {
   margin-bottom: 0.5rem;
 }
 .myGamesToggle {
-  label {
+  :is(label) {
     cursor: pointer;
   }
-  input {
+  :is(input) {
     cursor: pointer;
   }
 }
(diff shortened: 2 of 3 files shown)


benjaminpjones:
> That being said, I think that nested css syntax is quite handy in some cases (e.g. for deeply nested components)

Yeah I think this is why CSS pre-processors like [Sass](https://sass-lang.com/) are so popular.  Glad to see it moving into the standard!  I'm in favor of being able to use more modern CSS features.

The Vite recommendation seems to be to use a `PostCSS` plugin, which will translate modern css to older targets:

> Because Vite targets modern browsers only, it is recommended to use native CSS variables with PostCSS plugins that implement CSSWG drafts (e.g. [postcss-nesting](https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-nesting)) and author plain, future-standards-compliant CSS.

merowin:
> > That being said, I think that nested css syntax is quite handy in some cases (e.g. for deeply nested components)
> 
> Yeah I think this is why CSS pre-processors like [Sass](https://sass-lang.com/) are so popular. Glad to see it moving into the standard! I'm in favor of being able to use CSS nesting.
> 
> The Vite recommendation (see [docs](https://main.vitejs.dev/guide/features#css)) seems to be to use a `PostCSS` plugin, which will translate modern css to older targets:
> 
> > Because Vite targets modern browsers only, it is recommended to use native CSS variables with PostCSS plugins that implement CSSWG drafts (e.g. [postcss-nesting](https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-nesting)) and author plain, future-standards-compliant CSS.

I would like to try to add this plugin then, is that ok? Alternatively I could remove the nested css.

benjaminpjones:
Yes, I'm in favor!

merowin:
closing as this is obsolete with https://github.com/govariantsteam/govariants/pull/208

PR #199, opened 2024-02-17
195 server crashes on game request with wrong

### LINKED ISSUE #195: Server crashes on game request with wrong id
Steps to reproduce:

Navigate to a game view, change the game id in the URL and try to load the game view for this (not existing) game -> Server crashes.

Ideas to fix: catch the error, or include a check whether we actually found the game.

merowin:
fixes #195

pretty much just copied the code from playMove

Not sure if error code 500 is correct in this case, but this prevents the server from crashing.

### DIFF
--- a/packages/server/src/api.ts
+++ b/packages/server/src/api.ts
@@ -26,8 +26,13 @@ export const router = express.Router();
 
 // Set up express routes
 router.get("/games/:gameId", async (req, res) => {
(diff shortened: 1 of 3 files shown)


benjaminpjones:
> Not sure if error code 500 is correct in this case

Yeah I'm not sure it's the correct error code in the "play move" case either, but I'm not too concerned about it at this point

PR #204, opened 2024-02-19
remove npm from readme

merowin:
Removes the references to npm in readme file

The other change was automatically made, but I think its ok.


benjaminpjones:
I wonder if we should remove this README entirely, and move anything useful into the top-level docs.  This README is just instructions from Vue to new Vue users, not instructions from us to Go Variants contributors haha

PR #213, opened 2024-03-08
add sierpinsky triangle board

merowin:
Adds board pattern for sierpinsky triangle:
grafik

### DIFF
--- a/packages/shared/src/variants/badukWithAbstractBoard/abstractBoard/BadukBoardAbstract.ts
+++ b/packages/shared/src/variants/badukWithAbstractBoard/abstractBoard/BadukBoardAbstract.ts
@@ -6,6 +6,7 @@ import { CreatePolygonalBoard } from "./PolygonalBoardHelper";
 import { CreateCircularBoard } from "./CircularBoardHelper";
 import { Vector2D } from "./Vector2D";
 import { TrihexagonalBoardHelper } from "./TrihexagonalBoardHelper";
+import { createSierpinskyBoard } from "./SierpinskyBoard";
 
 export class BadukBoardAbstract implements IBadukBoard {
   Intersections: Intersection[];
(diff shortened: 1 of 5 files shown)


benjaminpjones:
This is WILD.. I love it! ☺️

Do we have a way to limit the sizes of these graph boards? We should probably have limits on all boards, but this one could be particularly problematic since the number of nodes is O(3^size) instead of O(size^2).

merowin:
Yeah I agree that a maximum size would be appropriate, but I'm not sure what should be the max.

I forgot to mention the thread: https://forums.online-go.com/t/sierpinsky-go/51048

benjaminpjones:
Thanks!

> I'm not sure what should be the max.

Up to you on the exact number.  4 is fine, and I think I'd be comfortable with as much as 6 or 7, since those are roughly equivalent to 19x19 and 33x33  in terms of nodes and edges.  For reference, `Baduk` goes up to 52x52.

| depth | side length | nodes | edges | grid approx
| --- | --- | --- | --- | --- |
| 1 | 4 | 3 | 3 | 2x2 |
| 2 | 8 | 6 | 9 | 3x2 |
| 3 | 16 | 15 | 27 | 4x4 |
| 4 | 32 | 42 | 81 | 7x7 |
| 5 | 64 | 123 | 243 | 11x11 |
| 6 | 128 | 366 | 729 | 19x19 |
| 7 | 256 | 1095 | 2187 | 33x33 |
| 8 | 512 | 3282 | 6561 | 57x57 |

merowin:
I chose the limit based on how tiny the distance between intersections get (currently). Here's some references:
Size 4
grafik

Size 5
grafik

Size 6
grafik

We can support these boards, as it doesn't take too long to render, but it seems difficult to play like this.
One "problem" of this board is that a lot of screen space is kind-of wasted.

On another note, I'm curious why the visual distance to the board edge is constant on top and bottom, but scales with the intersection distance left and right.

benjaminpjones:
> One "problem" of this board is that a lot of screen space is kind-of wasted.

Oh haha, I didn't think about the actual size of the board. You're absolutely right!

It's a nice demonstration of sub-2 Hausdorff dimension 😂

Issue #174, opened 2024-01-24
Some Moves Shouldn’t Be Sent to Clients


JonKo314:
Currently all moves of a game are sent to the clients via the `GET` `/games/:gameId` route. That's a problem in variants with parallel moves, because moves of an ongoing round should be secret.

benjaminpjones:
Agreed! See also https://github.com/govariantsteam/govariants/issues/48

merowin:
I think we generally need a plan for hidden information. For Parallel Variants it seems to be sufficient to remove the staged moves (except yours). For other hidden information games, maybe variants could be capeable of obfuscating moves, or similar.

merowin:
Should we close this?

benjaminpjones:
I think so. The only outstanding "todo" is that we haven't fully hidden the old APIs right?

merowin:
> I think so. The only outstanding "todo" is that we haven't fully hidden the old APIs right?

Yeah, and I guess that there is no server side validation of the socket subscriptions.

merowin:
I made a follow-up issue: https://github.com/govariantsteam/govariants/issues/301 and thus this can be closed.

Issue #48, opened 2022-08-29
State via the API


benjaminpjones:
In hidden information games, the clients should not have access to the whole moves list. My thought is that the following endpoints can be created:

- `/games/%id%/state`
- `/games/%id%/state/%seat%` (partial state)

It would return the output of `exportState()`.

benjaminpjones:
Can probably be done at the same time as #7

merowin:
My concern with this is the following. Currently the game view creates a whole game object, using all moves. That gives us a lot of flexibility; We can use the method nextToPlay, the gamePhase and result, the round method, etc.

One could argue that these things are (strictly speaking) part of what defines the game state. But currently we don't handle it that way. If we add all this to the exported game state, It might add a lot of complexity to the variants. And the api endpoint would be called a lot (e.g. when navigating through the moves).

benjaminpjones:
I agree with the fact that it is nice to have the game object in the client.  That said, I want to address this part specifically:

> If we add all this to the exported game state, It might add a lot of complexity to the variants.

Passing `nextToPlay`, `gamePhase` etc. need not complicate variant implementations.  We could also pass all these "common" attributes over the state API as well.  The payload could look something like:

```
{
  "variant_state": { "board": [...], captures: {...} },  // output of exportState()
  "next_to_play": [0],
  "phase": "gameover",
  "result": "B+R",
  "round": 103,
}
```

That said, there are always multiple approaches to a problem.  Is there a solution to "hidden information gets leaked via the API" that you see as preferable?

merowin:
I haven't worked out the details, but an alternative might be to allow variants to modify the move sequence before it is sent to clients.

Parallel move variants could remove the staged moves (except yours),
Hidden move variants could obfuscate moves

However, I'm not sure if its preferable. This approach might not give us the same flexibility that your approach does.

JonKo314:
One thing to consider: It might not always be possible to create the subjective game state by all moves that are shared with a user.

Example variant, like Baduk but:
- Players only see their own moves
- If they try to make an illegal move, they are prevented to make that move and may make a different move
- Players are told, how many stones they have captured at any given moment

merowin:
Yeah, you two convinced me. I'm on board with this 🙂

benjaminpjones:
Marking this high priority as there are other related issues (#266, #174), and the cost of this tech debt is grows quickly as other features are start to depend on the full game object existing client-side (#168, #214)

merowin:
[...]

PR #214, opened 2024-03-13
add button to submit moves

JonKo314:
Adds a client-side option to protect against misclicks by adding a move to the GameResponse before it's sent to the server. That's a bit problematic, because it leads to side effects which need special treatment, as the usage of the `resetMoveToSubmit()` method shows.

It might be better to do this with the board components, because those have to display the move, but then we'd have to implement it for different variants.

### DIFF
--- a/packages/vue-client/src/views/GameView.vue
+++ b/packages/vue-client/src/views/GameView.vue
@@ -13,7 +13,7 @@ import type {
   IPerPlayerTimeControlBase,
   IConfigWithTimeControl,
 } from "@ogfcommunity/variants-shared";
-import { computed, reactive, ref, watchEffect, type Ref } from "vue";
+import { computed, reactive, ref, watchEffect, watch, type Ref } from "vue";
 import { board_map } from "@/board_map";
 import { socket } from "../requests";
 import { variant_short_description_map } from "../components/variant_descriptions/variant_description.consts";
@@ -31,6 +31,10 @@ const DEFAULT_GAME: GameResponse = {
 // while viewing history of game, maybe we should prevent player from making a move (accidentally)
 const view_round: Ref<number | null> = ref(null);
 const navigateRounds = (offset: number): void => {
+  if (resetMoveToSubmit()) {
+    return;
+  }
+
   if (game.value.round === undefined) {
     // game not initialised
     return;
@@ -141,9 +145,17 @@ const setPlayingAs = (seat: number) => {
   }
 };
 
-function makeMove(move_str: string) {
-  const move: { [player: number]: string } = {};
+const moveToSubmit = ref<{ [player: number]: string } | null>(null);
+const useSubmitButton = ref<boolean>(
+  (
+    localStorage?.getItem("settings.useSubmitButton") ?? "false"
+  ).toLowerCase() === "true",
+);
+watch(useSubmitButton, () => {
+  localStorage?.setItem("settings.useSubmitButton", `${useSubmitButton.value}`);
+});
 
+function makeMove(move_str: string) {
   if (playing_as.value === undefined) {
     alert("No player selected. Click one of your seats to play a move.");
     return;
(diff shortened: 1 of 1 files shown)


merowin:
Nice, looks good to me!

> That's a bit problematic, because it leads to side effects which need special treatment, as the usage of the resetMoveToSubmit() method shows.

One possibility to consider may be to not push the moveToSubmit into gameResponse.moves, but then add some logic to the computed function of game. That may allow us to avoid certain UI aspects being updated (e.g. game result, next to play indicator) if we want that.
Not sure if that helps with the reset function though :D

merowin (approved):

benjaminpjones (approved):
Awesome! Yeah this seems super important for correspondence.

I agree pushing into the GameResponse seems weird, but I'm fine with it.  The main concern I have is that this is one more feature we will need to re-think when we move the moves logic server-side.  (#174, #48)

merowin:
That's a good point. What we need here is like a "move preview", and taking the normal move transition is kind-of conflicting with hidden information variants, where moves might reveal new information.

JonKo314:
> That's a good point. What we need here is like a "move preview", and taking the normal move transition is kind-of conflicting with hidden information variants, where moves might reveal new information.

Yeah, but the hidden information should never be shared with the client in the first place, so I think that's another issue. I agree about the "move preview" though.

merowin:
I agree, I was just thinking that for variants with hidden information, we can't use the normal move transition as a move preview.

But all that shouldn't hold this back, I think.

benjaminpjones:
I think this actually does make it a lot easier to reveal moves early. For parallel games, if someone has a setting on and ends the round, they can see opponent moves without submitting.

Granted, you can already do this by looking at your network tab. But yeah I think if we could land on a solution that doesn't involve moves, it will make our lives easier when it we push this stuff to the server.

JonKo314:
> I think this actually does make it a lot easier to cheat. For parallel games, if someone has a setting on and ends the round, they can see opponent moves without submitting.

And it could put users into an uncomfortable position, even without the intent to cheat.

So now I think it would be better not to merge this currently.

benjaminpjones:
Converted to draft given the concern for simultaneous variants.

Happy to discuss design - these are what I have in mind, but probably not the only options:

- Place a dummy stone or marker on the board, but don't do any game logic <-- my preference
   - Pro: Can be used for hidden info games
   - Con: Somewhat reduced experience for full-info games as we won't see captures previewed
   - Con: needs to be implemented for each board
- Keep this implementation, but turn it off for hidden info
   - Pro: full experience for full-info games
   - Con: no feature for hidden info/simul
- Hybrid of above (use moves for full-info, don't use moves for hidden info)
   - Possibly best user experience
   - Con: needs to be implemented for each board
   - Con: maintenance burden 😞

JonKo314:
Another option would be to apply the game logic to the information the client is allowed to know about.

If a player wants to capture a stone by playing K10, but there is already a stone there which is unknown to the player, their new stone will be shown at K10 and the stone that is about to be captured disappears. Then the player submits K10 and only after that the collision is handled according to the game rules.

- Pro: Full experience for perfect information games
- Pro: Some experience for games with imperfect information
- Con: Requires a somewhat predictable game state even with imperfect information (maybe acting as if there was no hidden information isn't easy in some future variants)

merowin:
I like this idea. Given (#48), maybe we can make it so a new instance of AbstractGame can be created from the game state. For hidden information games, that may be a partial state of sorts, and we can think of the move transition from a partial state instance as a move preview.

benjaminpjones:
I agree this has potential to give the user the highest quality experience.  A couple of thoughts - nothing blocking:

> Given (https://github.com/govariantsteam/govariants/issues/48), maybe we can make it so a new instance of AbstractGame can be created from the game state.

We actually had an `importState` function originally, but [I removed it](https://github.com/govariantsteam/govariants/pull/95/files) because it wasn't used and only implemented for some variants. Will we want to make this function required?  And how can we convey to variant implementers what is expected of this function?

> For hidden information games, that may be a partial state of sorts

`exportState()` isn't even a complete representation for perfect info variants!  For example, we leave out [`SuperkoDetector`](https://github.com/govariantsteam/govariants/blob/7dbf9857e52442375a7faf9aa3909aab24c89c77/packages/shared/src/variants/baduk.ts#L31) because it's a larger object and not very important for displaying the board.

JonKo314:
> It would be kind of strange to preview the capture, then undo it after clicking the button. Of course this can be updated to give the preview, but it does make me wonder if this would be an oversized burden for variant implementers.

Good point. I didn't really consider Fractional when composing the example above. It feels natural not to show the stone as captured in this case, because the moves of the other players aren't decided yet. So nothing happens until all moves of the current round are known, just as the game logic does too.

We probably also don't want to show previews for one color go.

benjaminpjones:
Closing due to conflicts with the new "game state over network" architecture (#48)

PR #215, opened 2024-03-23
Remove timeout button, Some time handling fixes

merowin:
1. removes the timeout button (my bad)
2. pass is not allowed in capture variant
3. fix some issues with time handling:
Game ending moves need to be handled, otherwise the winning players time keeps running.
If A player times out, its time is set to 0, so it is displayed correctly.

### DIFF
--- a/packages/server/src/games.ts
+++ b/packages/server/src/games.ts
@@ -151,20 +151,19 @@ export async function handleMoveAndTime(
   ) {
     if (game_obj.phase === "gameover") {
       getTimeoutService().clearGameTimeouts(game.id);
-    } else {
-      if (Object.keys(timeControlHandlerMap).includes(game.variant)) {
-        const timeHandler = new timeControlHandlerMap[game.variant](
-          new Clock(),
-          getTimeoutService(),
-        );
-        timeControl = timeHandler.handleMove(
-          game,
-          game_obj,
-          playerNr,
-          new_move,
-          isRoundTransition,
-        );
-      }
+    }
+    if (Object.keys(timeControlHandlerMap).includes(game.variant)) {
+      const timeHandler = new timeControlHandlerMap[game.variant](
+        new Clock(),
+        getTimeoutService(),
+      );
+      timeControl = timeHandler.handleMove(
+        game,
+        game_obj,
+        playerNr,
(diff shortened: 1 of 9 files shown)


benjaminpjones:
> pass is not allowed in capture variant

Why not? In the Capture Go tournaments I've played on OGS, games were allowed to go to scoring (for which you need pass)

For example, from the description in [Capture Go 10th Cycle](https://online-go.com/tournament/55560)

> If no stones are captured during the game, then the game is scored like a normal Japanese game.

merowin:
> > pass is not allowed in capture variant
> 
> Why not? In the Capture Go tournaments I've played on OGS, games were allowed to go to scoring (for which you need pass)
> 
> For example, from the description in [Capture Go 10th Cycle](https://online-go.com/tournament/55560)
> 
> > If no stones are captured during the game, then the game is scored like a normal Japanese game.

Hm I didn't know that this variation of the rules exists. I was thinking of the variation where capturing a stone is the ultimate goal, and passing is not allowed (otherwise passing from the start can't be beat). I.e. [https://forums.online-go.com/t/anyone-interested-in-playing-capture-go/17206](url)

So should I undo that change?

benjaminpjones:
> https://forums.online-go.com/t/anyone-interested-in-playing-capture-go/17206

Oh haha yeah they do discuss this ruleset and the tournaments at that thread

> So should I undo that change?

I'd like to keep passes around in some form, given the popularity of this version of the rules in the OGS tournaments. I'm fine if you make it configurable though.

Could also just be a separate PR since it's unrelated to time control.

merowin:
> > https://forums.online-go.com/t/anyone-interested-in-playing-capture-go/17206
> 
> Oh haha yeah they do discuss this ruleset and the tournaments at that thread
> 
> > So should I undo that change?
> 
> I'd like to keep passes around in some form, given the popularity of this version of the rules in the OGS tournaments. I'm fine if you make it configurable though.
> 
> Could also just be a separate PR since it's unrelated to time control.

Right, makes sense. I've reverted the change that removed passes in capture.

I had thought that this was an oversight, but now that I think about it, this makes a lot of sense, so players need not fill in their own territory at the end.

PR #217, opened 2024-03-27
Quantum fix

benjaminpjones:
From Zi

> We did find a minor issue, i.e., the quantum pairs should live or die together, so once they are removed, the paired positions are still paired, meaning if a new stone is placed on one board in one of the quantum positions, it should also appear in the paired position on the other board.

image(1)
image(2)

---

This update actually simplifies the code because we no longer need to break entanglements.

- Move all quantum logic out of the helper class (and rename to `BadukHelper`)
- Represent quantum stones with a 1D 2-element array, instead of a 2D array
- Fix the quantum "reviving" of stones
- Add two tests

### DIFF
--- a/packages/shared/src/variants/__tests__/quantum.test.ts
+++ b/packages/shared/src/variants/__tests__/quantum.test.ts
@@ -236,3 +236,46 @@ test("Placing a stone in a captured quantum position", () => {
 
   expect(game.exportState().quantum_stones).toEqual(["ia", "ai"]);
 });
+
+test("Placing a white stone in a captured quantum position", () => {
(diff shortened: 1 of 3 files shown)


benjaminpjones:
I'm going to land this one without review, as I think the only way Zi and team can review is if it's deployed :/

merowin:
> I'm going to land this one without review, as I think the only way Zi and team can review is if it's deployed :/

In my opinion its not too difficult to get the project running locally, if they're willing to put in a bit of effort, and with a bit of help.

I'm fine that you deployed this, but I'm hoping that Zi and team help us with the development. Getting the project to run locally would be a first step.

PR #223, opened 2024-03-29
Add Byo-Yomi, Canadian, and Simple time controls

benjaminpjones:
Update: I've plugged it in, and both the old and new time controls seem to work as expected!  Good test suite on the handlers too - helped me debug a lot of issues on my first pass.

Sorry it's such a big diff.  Hopefully the break down of the commits help to understand what's going on.  I'm also going to leave this open for a while so I can test some more, but publishing now because it seems ready.

---

This is a proposal for a generic interface for additional time controls.  I have not plugged it into the code yet, but I hope it demonstrates a strategy that we can use to onboard Byoyomi, Simple and Canadian to reach parity with OGS.

I am opening this PR as draft to make sure I don't collide with anyone and also gather early feedback, but I plan to do the following before marking "Ready":

- [x] plug this into the code
- [x] expose the new time controls in the UI
- [x] Add config forms
- [x] Write unit tests for each of the `TimeControlSystem`s

## Interface

`TimeControl` has the following functions:

- `initialState`
    - corresponds to code in [TimeHandlerParallelMoves.initialState](https://github.com/govariantsteam/govariants/blob/byoyomi/packages/server/src/time-control/time-handler-parallel.ts#L37-L41) and [TimeHandlerSequentialMoves.initialState](https://github.com/govariantsteam/govariants/blob/byoyomi/packages/server/src/time-control/time-handler-sequential.ts#L36-L39)
- `elapse` and `renew`
    - correspond roughly to the `transition` functions in [TimeHandlerParallelMoves](https://github.com/govariantsteam/govariants/blob/7dbf9857e52442375a7faf9aa3909aab24c89c77/packages/server/src/time-control/time-handler-parallel.ts#L82-L110) and [TimeHandlerSequentialMoves](https://github.com/govariantsteam/govariants/blob/7dbf9857e52442375a7faf9aa3909aab24c89c77/packages/server/src/time-control/time-handler-sequential.ts#L77-L116)
    - I split into two functions because we can use `elapse` to generalize the [subtraction in the client countdown timer](https://github.com/govariantsteam/govariants/blob/7dbf9857e52442375a7faf9aa3909aab24c89c77/packages/vue-client/src/components/GameTimer.vue#L84C1-L84C67)
- `msUntilTimeout`
   - corresponds to [TimeHandlerParallelMoves.getMsUntilTimeout](https://github.com/govariantsteam/govariants/blob/7dbf9857e52442375a7faf9aa3909aab24c89c77/packages/server/src/time-control/time-handler-parallel.ts#L150) and [TimeHandlerSequentialMoves.getMsUntilTimeout](https://github.com/govariantsteam/govariants/blob/7dbf9857e52442375a7faf9aa3909aab24c89c77/packages/server/src/time-control/time-handler-sequential.ts#L150)
- ~~`toString`~~`timeString`
    - for the client code ([GameTimer.vue](https://github.com/govariantsteam/govariants/blob/7dbf9857e52442375a7faf9aa3909aab24c89c77/packages/vue-client/src/components/GameTimer.vue#L24))

## Demo

https://github.com/govariantsteam/govariants/assets/25233703/604982b4-94cd-4e09-ba89-1bcd9a39e7a1

### DIFF
--- a/packages/server/src/games.ts
+++ b/packages/server/src/games.ts
@@ -71,6 +71,20 @@ export async function getGame(id: string): Promise<GameResponse> {
   if (!game.players) {
     game.players = await BACKFILL_addEmptyPlayersArray(game);
   }
+  const playerClocks = Object.values(game.time_control?.forPlayer ?? {});
+  // bpj 2024-04-04: changed playerClock.remainingTimeMS to playerClock.clockState
+  // in order to support more complex clock states such as byo-yomi
+  if (game.time_control && playerClocks.some((tc) => tc.clockState == null)) {
+    playerClocks.forEach((tc) => {
+      if (tc.clockState) return;
+
+      if ("remainingTimeMS" in tc) {
+        // This is the state for Fischer and Absolute, the only supported
+        // time controls at the time of this change.
+        tc.clockState = { remainingTimeMS: tc.remainingTimeMS };
+      }
+    });
+  }
 
   return game;
 }
--- a/packages/server/src/time-control/time-handler-parallel.ts
+++ b/packages/server/src/time-control/time-handler-parallel.ts
@@ -100,7 +100,10 @@ export class TimeHandlerParallelMoves implements ITimeHandler {
 
     const playerData = timeControl.forPlayer[playerNr];
 
-    const timeConfig = (game.config as IConfigWithTimeControl).time_control;
+    if (!game.config.time_control) {
+      throw new Error(`Time control not found in game ${game.id}`);
+    }
+    const timeConfig = game.config.time_control;
     const clockController = timeControlMap.get(timeConfig.type);
 
     if (move === "resign" || move === "timeout") {
--- a/packages/server/src/time-control/time-handler-parallel.ts
+++ b/packages/server/src/time-control/time-handler-parallel.ts
@@ -5,6 +5,7 @@ import {
   ITimeControlBase,
   PerPlayerTimeControlParallel,
   TimeControlParallel,
+  timeControlMap,
 } from "@ogfcommunity/variants-shared";
 import { ITimeoutService } from "./timeout";
 import { ITimeHandler } from "./time-control";
@@ -13,6 +14,7 @@ import {
   getMsUntilTimeout,
   initialState,
   makeTransition,
+  nullTimeState,
 } from "./time-handler-utils";
 
 export class TimeHandlerParallelMoves implements ITimeHandler {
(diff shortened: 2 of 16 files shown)


packages/shared/src/time_control/time_control_systems.ts
```diff
@@ -0,0 +1,258 @@
...
+    state = { ...state };
+    if (state.mainTimeRemainingMS) {
+      if (state.mainTimeRemainingMS >= elapsedMS) {
+        state.mainTimeRemainingMS -= elapsedMS;
+        return state;
+      }
+      elapsedMS -= state.mainTimeRemainingMS;
+      state.mainTimeRemainingMS = 0;
+    }
+    while (state.periodTimeRemainingMS < elapsedMS) {
```
merowin, line 109: My instinct would be to replace this while by division with remainder. Something like
```
let periodsUsed = 0;

// first period
if (state.periodTimeRemainingMS < elapsedMS) {
  elapsedMS -= state.periodTimeRemainingMS;
  periodsUsed += 1;
}

let remainder = elapsedMS % config.periodTimeMS;
const q = (elapsedMS - remainder) / config.periodTimeMS; // should give integer result
periodsUsed += q;

state.numPeriods = Math.max(0, state.numPeriods - periodsUsed);
state.periodTimeRemainingMS = state.numPeriods === 0
  ? 0
  : config.periodTimeMS - remainder;
```
I haven't tried this out yet, and actually, I suspect your solution has better performance (as long as there are not many periods used in one move, which can't happen often anyways).

packages/shared/src/time_control/time_control_systems.ts
```diff
@@ -0,0 +1,258 @@
...
+    }`;
+    if (state.mainTimeRemainingMS > 0) {
+      return `${msToTime(state.mainTimeRemainingMS)} + ${periodTimeString}`;
+    }
+    return periodTimeString;
+  }
+}
+
+// Convenience function for when the state doesn't change at the end of a round
+function nullRenewal<T>(state: T) {
```
merowin, line 245: nit: I believe T needs to be an iterable for this to work (or?). We could e.g. say
`nullRenewal<T extends object>`

merowin (approved):
Awesome, looks great!

packages/server/src/time-control/time-handler-parallel.ts
```diff
@@ -98,20 +100,22 @@ export class TimeHandlerParallelMoves implements ITimeHandler {
 
     const playerData = timeControl.forPlayer[playerNr];
 
+    const timeConfig = (game.config as IConfigWithTimeControl).time_control;
+    const clockController = timeControlMap.get(timeConfig.type);
```
merowin, line 104: should we guard against the possibility that this returns undefined?

benjaminpjones (commented):
Thanks for the review @merowin!

packages/server/src/games.ts
```diff
@@ -71,6 +71,20 @@ export async function getGame(id: string): Promise<GameResponse> {
   if (!game.players) {
     game.players = await BACKFILL_addEmptyPlayersArray(game);
   }
+  const playerClocks = Object.values(game.time_control?.forPlayer ?? {});
+  // bpj 2024-04-04: changed playerClock.remainingTimeMS to playerClock.clockState
```
benjaminpjones, line 75: I just tested this part manually by starting a long Fischer game, editing the fields to look like an old game, then playing some moves.  It works fine, but leaves the old field behind, and never updates it.

Screenshot 2024-04-05 at 9 15 04 PM

It's not concerning enough for me to try and mess with it more in this PR, but something to think about in the discussion in #225.

PR #227, opened 2024-03-31
Dedup handlers

benjaminpjones:
#223 is going to have a lot of extra code if we don't clean up first.  The two handlers share a lot of code, this diff moves some of that code into shared functions.

### DIFF
--- a/packages/server/src/time-control/time-handler-utils.ts
+++ b/packages/server/src/time-control/time-handler-utils.ts
@@ -1,0 +1,57 @@
+import {
+  GenericTimeControl,
+  IConfigWithTimeControl,
+  IPerPlayerTimeControlBase,
+  ITimeControlBase,
+  TimeControlType,
+  makeGameObject,
+} from "@ogfcommunity/variants-shared";
+
+export function initialState(
+  variant: string,
+  config: IConfigWithTimeControl,
+): ITimeControlBase;
+export function initialState<T extends IPerPlayerTimeControlBase>(
+  variant: string,
+  config: IConfigWithTimeControl,
+  modifier: (per_player_base: IPerPlayerTimeControlBase) => T,
+): GenericTimeControl<T>;
+export function initialState(
+  variant: string,
+  config: IConfigWithTimeControl,
+  modifier: (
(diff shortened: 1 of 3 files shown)


merowin (approved):
Looks great and much more concise! Now I feel a bit bad that I left behind such a mess ;-)

packages/server/src/time-control/time-handler-utils.ts
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        playerData.remainingTimeMS -= getElapsedMS(playerData);
+      };
+    }
+
+    case TimeControlType.Fischer: {
+      const fischerConfig = config.time_control as IFischerConfig;
+      return (playerData) => {
+        if (move === "timeout") {
```
merowin, line 82: Doesn't necessarily concern this PR, but I just had a thought. Currently the timeout handler triggers a timeout move, and then calls the time handler again. I believe this makes sense, because e.g. for parallel games, the timeout may cause the round to end, and the time handler needs to do stuff.
However it's kind of annoying that each single-transition needs to handle this case. Should I try to refactor the timeout handler s.t. it updates the remaining time itself, and maybe then this isn't necessary?

benjaminpjones: There may be room for optimization in whether the handler needs to be called twice, but we can also prevent the problem that "each single-transition needs to handle this case" by moving the logic outside of the `switch` statement...  or removing the `switch` entirely in favor of polymorphic approach.  Here is the way I plan to reword this function in #233:

```
export function makeTransition(...) {
    const timeConfig:  = config.time_control;
    const control: TimeControlSystem = timeControlSystemMap.get(time_config.type);
    
    return (playerData) => {
        const timeState = playerData.timeState;
        const elapsedMS = getElapsedMS(playerData);
        if (move === "timeout") {
          playerData.remainingTimeMS = 0;
          return;
        }
        const elapsed = control.elapse(elapsedMS, timeState, timeConfig);
        playerData.timeState = control.renew(timeState, timeConfig);
    };
}
```

PR #228, opened 2024-03-31
Clean up the GameTimer Vue component

benjaminpjones:
The logic gets a lot simpler if we can assume time_control is not null.  Anyway, `<GameTimer time_control="null">` isn't very meaningful.  So instead we type it as non-null and guard it with a v-if.

Some other small changes: make more use of computed variables, and try to remove variables while keeping the same functionality.

Motivation: this should help a bit with #233


merowin (approved):
This simplifies the timer component a lot, thank you!

Issue #231, opened 2024-04-06
Timeout bugs in parallel variant


benjaminpjones:
Scenario: 4 players, two players time out in the same round -> round freezes

- if "active" players try to play, they either change their move or get the "not enough time to change move" message
- if timed-out players try to play, they get "not your turn"

Scenario: 4 players, one player times out -> game ends and says Player 0 wins!

merowin:
About the Scenario with 4 players, one times out <- I've tested this in the past, and it used to work. I'm not sure what exactly is going wrong, it looks like all 4 players time out at the same time for some reason.

merowin:
I have found out the reason for this bug. The new function MakeTransition takes the move as input parameter. However in parallel time control, a move from any player may trigger a round transition. At this point we don't know the moves of other players, and it seems that this move is used for the transition for all players. In the event that the move is "timeout", the transition sets all player times to 0. Then all players immediately time out.

merowin:
I'm not sure how to fix it without disrupting your bigger PR.

benjaminpjones:
Ohh oh no! Thank you for root causing. And sorry for breaking it 😬

> I'm not sure how to fix it without disrupting your bigger PR.

Should I go ahead and land it then? Or I am also okay with dealing with some merge conflicts if we need to fix before landing

merowin:
Please go ahead and land the PR. All in all, I think It'll be easier to fix it afterwards.

> And sorry for breaking it 😬

No problem! 😆

benjaminpjones:
I believe the second issue "two players time out in the same round" is a race condition.  It is very easy to trigger if it happens in the first timed round, but otherwise shouldn't happen often.

To repro:

1) Set up 4 player parallel game (or any parallel variant with >2 players).
2) Finish round 1 to start the clock
3) play moves for player 0 and player 1
4) Wait for player 2 and player 3 to timeout

Results:

It appears that the timeout for player 3 is overwritten or never occurs.  If there is a race, it means that neither timeout will detect that the round ends.

Timeout in round two, simultaneous, BAD:

```
    "forPlayer": {
      "0": {
        "clockState": {
          "remainingTimeMS": 30000
        },
        "onThePlaySince": "2024-04-14T20:29:43.452Z",
        "stagedMoveAt": "2024-04-14T20:30:03.660Z"
      },
      "1": {
        "clockState": {
          "remainingTimeMS": 30000
        },
[...]

PR #232, opened 2024-04-06
fix transition for parallel time handler

merowin:
issue: https://github.com/govariantsteam/govariants/issues/231

This should fix the basic timeout scenario for parallel move variants. The scenario where multiple people timeout at almost the same time does not get fixed by this.

### Changes
Remove the move from the transition. Due to how the parallel move time handler works currently (we don't consider the moves except for the move that triggers the round transition), we can't handle specific moves like timeout in the transition.

Remark: The deleted code parts are not important for the timeout scheduling (as a timeout move was already played, when this is relevant).

I've also added a check in the parallel handler to ensure that stagedMoveAt is set when transition is called.

### DIFF
--- a/packages/server/src/time-control/__tests__/time-handler-utils.test.ts
+++ b/packages/server/src/time-control/__tests__/time-handler-utils.test.ts
@@ -26,12 +26,7 @@ test("initialState (Absolute)", () => {
 });
 
 test("makeTransition (Absolute)", () => {
-  const transition = makeTransition(
-    () => 5000,
-    ABSOLUTE_CONFIG,
-    "aa",
-    "game_id",
-  );
+  const transition = makeTransition(() => 5000, ABSOLUTE_CONFIG, "game_id");
   const playerData: IPerPlayerTimeControlBase = {
     clockState: { remainingTimeMS: 600000 },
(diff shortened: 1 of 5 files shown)


merowin:
I've "fixed" the test cases by commenting out those that fail 😅 

Unfortunately I believe that the current structure of the parallel moves handler makes it fairly difficult to handle the transition for timeout moves elegantly.

benjaminpjones:
Thanks for fixing this! I think [your comment](https://github.com/govariantsteam/govariants/pull/227#discussion_r1548329909) on the other PR was related to this, but I didn't see the change in functionality

> I've "fixed" the test cases by commenting out those that fail 😅

Totally fine delete tests that don't make sense any more - I am not attached to any specific test 😁 ideally, we will find a way to test the 2 scenarios from my bug report. I have a draft with the setup code, but it's not failing in the way I'd expect based on manual tests.

merowin:
I've made it so the timeout handler sets the null clock state after timeout. I think this makes sense both from a thematical standpoint, and because we handle this at one central place.

I didn't guard against the force-unwrapping in any way, since it is inside a try-catch anyway.

PR #233, opened 2024-04-06
[don't land] An attempt at a test for #231

benjaminpjones:
I am trying to reproduce the "1 player timing out of 4 player game ends the game" bug in a unit test.  Not quite there, but here's what I have so far

I marked this as "don't land" since the test should be something that "fails" on `main`.  So far, it looks like no bad scheduled timeouts and the state is as-expected.


benjaminpjones:
Closing, since this test doesn't cover the bugs in #231, and a proper test would need to exercise scheduleTimeout and/or handleMoveAndTime

Issue #28, opened 2022-07-13
[dev] npm start doesn't seem to update shared


benjaminpjones:
I'm not sure how to fix this.  Maybe by importing the ts files into client/server directly rather than pointing to the dist dir.

merowin:
Would it be possible to add a script that first builds shared, and then builds + starts the server / vue-client?

JonKo314:
I think there should be some yarn setup that just makes this work with workspaces, but I haven't figured out how to do it yet.

benjaminpjones:
We are already using `-t`/`--topological` flag in the build script (see [workspaces foreach](https://yarnpkg.com/cli/workspaces/foreach) docs), which (I think) ensures `shared:build` gets called first.  It would be straightforward to roll `build` into the `start` command.

```
"start": "yarn build && yarn workspaces foreach -pi run start",
```

**However**, I don't think that would be sufficient to resolve this issue, since we want both servers to update on code changes in `shared`.  I think we might need something additional config for the watcher commands (`nodemon` for server, `vite` for vue-client).  And maybe some way to bypass the build command in shared, and let `vue-client` and `server` compile the `shared` TS.

benjaminpjones:
https://github.com/vitejs/vite/issues/819

For the Vite side, I think there is a solution in this thread

PR #240, opened 2024-04-19
Hot Reload Vue on shared/ changes

benjaminpjones:
Trying to figure out #28 (hot reload when shared/ changes)

While this was the one way I found to get hot reload to work, I don't think this solution is great for a couple of reasons.

- `"@/../../shared"` - it's pretty hacky to use relative imports.  It means we'll need to change a ton of files if we change the directory structure.
   - I'd prefer to import from `"@shared"` or something
   - I tried fiddling with "paths" and other similar config options, but no luck
- Would be better if we could depend on `dist` instead of `src`
    - I tried doing some of the magic using `watch.ignored` and `optimizeDeps` (see https://github.com/govariantsteam/govariants/tree/attempt-at-hot-reload) but wasn't able to turn on hot reload

That said, I am well beyond the limit of my TS and Vue/Vite knowledge, I am hoping someone might see a better path now that we have a "working" path

## Hot Reload Demo

Notice I am editing a variable in `shared/src/index.ts`!!!

https://github.com/govariantsteam/govariants/assets/25233703/8006b598-d154-4ac6-aee5-55581feb4762

### DIFF
--- a/packages/vue-client/src/views/HomeView.vue
+++ b/packages/vue-client/src/views/HomeView.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import GameCreationForm from "@/components/GameCreation/GameCreationForm.vue";
 import GameList from "@/components/GameList.vue";
-import { SITE_NAME } from "@/../../shared/src";
+import { SITE_NAME } from "@shared/index";
 </script>
 
 <template>
--- a/packages/vue-client/src/views/HomeView.vue
+++ b/packages/vue-client/src/views/HomeView.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import GameCreationForm from "@/components/GameCreation/GameCreationForm.vue";
 import GameList from "@/components/GameList.vue";
-import { SITE_NAME } from "@shared";
+import { SITE_NAME } from "@ogfcommunity/variants-shared";
 </script>
 
 <template>
--- a/packages/vue-client/src/views/HomeView.vue
+++ b/packages/vue-client/src/views/HomeView.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import GameCreationForm from "@/components/GameCreation/GameCreationForm.vue";
 import GameList from "@/components/GameList.vue";
-import { SITE_NAME } from "@shared/index";
+import { SITE_NAME } from "@shared";
 </script>
 
 <template>
--- a/packages/vue-client/src/views/HomeView.vue
+++ b/packages/vue-client/src/views/HomeView.vue
@@ -1,11 +1,12 @@
 <script setup lang="ts">
 import GameCreationForm from "@/components/GameCreation/GameCreationForm.vue";
 import GameList from "@/components/GameList.vue";
+import { SITE_NAME } from "@/../../shared/src";
 </script>
 
 <template>
   <main>
-    <h2>Welcome to Go variants!</h2>
+    <h2>Welcome to {{ SITE_NAME }}!</h2>
     <div>
       <GameCreationForm />
       <hr />
--- a/packages/vue-client/tsconfig.app.json
+++ b/packages/vue-client/tsconfig.app.json
@@ -7,10 +7,12 @@
     "baseUrl": "..",
     "rootDir": "..",
     "paths": {
-      "@/*": ["vue-client/src/*"], 
+      "@/*": ["vue-client/src/*"],
+      "@shared": ["./shared/src/index"],
       "@shared/*": ["./shared/src/*"]
     }
     ,
-    "verbatimModuleSyntax": false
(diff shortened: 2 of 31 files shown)


JonKo314:
> * it's pretty hacky to use relative imports
Yeah

> * Would be better if we could depend on `dist` instead of `src`
I'm not so sure about that one. `src` has all the typed stuff and doesn't need to be compiled, but it's just a feeling, I don't know enough about it.

It's very cool that you got it to work, but it would be sad if there wasn't a nicer way that just works.

benjaminpjones:
> > Would be better if we could depend on dist instead of src
> 
> I'm not so sure about that one. src has all the typed stuff and doesn't need to be compiled, but it's just a feeling, I don't know enough about it.

So I don't feel too strongly about this point compared to the first one.  However, my thought process is that `dist` is the directory that consumers would normally use when importing from the package (`"@ogfcommunity/variants-shared"`).

> it would be sad if there wasn't a nicer way that just works.

Indeed!  I'll need some help though - I've tried quite a few ways with very little luck 😞

benjaminpjones:
Okay actually, I think I got the absolute imports figured out.  Using `./shared` instead of `shared` in compilerOptions.paths helps

benjaminpjones:
Okay, I think I actually got this in a place that's publishable.

Summary of changes since earlier conversation:

- figured out how to keep the imports the same as before (not relative)
- I changed back to @ogfcommunity/variants-shared from @shared
    - not because I'm married to that name, but I wanted to keep focus on config changes in this PR
- Reverted tsconfig.vitest.json because it's of no consequence

packages/vue-client/src/views/HomeView.vue
```diff
@@ -1,11 +1,12 @@
 <script setup lang="ts">
 import GameCreationForm from "@/components/GameCreation/GameCreationForm.vue";
 import GameList from "@/components/GameList.vue";
+import { SITE_NAME } from "@ogfcommunity/variants-shared";
 </script>
 
 <template>
   <main>
-    <h2>Welcome to Go variants!</h2>
+    <h2>Welcome to {{ SITE_NAME }}!</h2>
```
benjaminpjones, line 9: This was mainly for testing.  I can remove it before merging, but want to make sure we landed on an agreeable configuration first.

packages/vue-client/tsconfig.app.json
```diff
@@ -4,9 +4,14 @@
...
-    "baseUrl": ".",
+    "baseUrl": "..",
+    "rootDir": "..",
     "paths": {
-      "@/*": ["./src/*"]
+      "@/*": ["vue-client/src/*"],
+      "@ogfcommunity/variants-shared": ["./shared/src/index"]
     }
+    ,
+    "verbatimModuleSyntax": false,
```
benjaminpjones, line 14: I think `verbatimModuleSyntax` off makes it so I don't need to put `type { SomeImport }` **everywhere** in `shared/`.  I'm not opposed to it, but it would be quite a large change, and I wanted to keep this diff to config-only as much as possible.

packages/vue-client/tsconfig.app.json
```diff
@@ -4,9 +4,14 @@
   "exclude": ["src/**/__tests__/*"],
   "compilerOptions": {
     "composite": true,
-    "baseUrl": ".",
+    "baseUrl": "..",
```
benjaminpjones, line 7: This puts us in the `packages` directory, so we can look at `packages/shared`

packages/vue-client/tsconfig.config.json
```diff
@@ -3,6 +3,14 @@
   "include": ["vite.config.*", "vitest.config.*", "cypress.config.*"],
```
benjaminpjones, line 3: I'm pretty sure the changes were required in all 3 tsconfig files (excluding tsconfig.vitest.json). Maybe there is a way to dedupe using the "extends" field.  But I don't quite understand the difference in intent behind `app` and `config` anyway...  Maybe we can remove them and just keep tsconfig.json?

packages/vue-client/vite.config.ts
```diff
@@ -9,6 +9,12 @@ export default defineConfig({
   resolve: {
     alias: {
       "@": fileURLToPath(new URL("./src", import.meta.url)),
+      "@ogfcommunity/variants-shared/src": fileURLToPath(
+        new URL("../shared/src", import.meta.url),
```
benjaminpjones, line 13: `@ogfcommunity/variants-shared/src` is to support imports like "@ogfcommunity/variants-shared/src/time_control"

`@ogfcommunity/variants-shared/src` is to support imports like "@ogfcommunity/variants-shared".

Importantly, the /src version needs to come first, else the latter will overtake it.

PR #246, opened 2024-04-24
graph baduk

merowin:
Not completely functional yet, work in progress.


benjaminpjones:
Looks cool.. but I don't quite understand, is this just a re-implementation of "badukWithAbstractBoard"?

merowin:
> Looks cool.. but I don't quite understand, is this just a re-implementation of "badukWithAbstractBoard"?

Kindof, the aim is also to create a variant that is suitable to be a base class for other variants that are playable on graph board patterns (similar to how Baduk is a base class for many variants).

This rework will also reduce boilerplate code a bit (only one place where new board patterns need to be added - the board factory).
I'd also like to add scoring, which badukWithAbstractBoard is currently lacking.

merowin:
I'm closing this because my plans have changed 😅

PR #247, opened 2024-04-26
Split out NavButtons into a separate component

benjaminpjones:
## Motivation

Modularity is generally a good thing, and I am specifically looking to reduce code duplication as I create a separate "client-only" game page (check out the [variant-demo branch](https://github.com/govariantsteam/govariants/tree/variant-demo) if you're interested!).

## Changes in this PR

- Upgrade Vue: 3.4 has a nice function [`defineModel()`](https://vuejs.org/api/sfc-script-setup.html#definemodel)
- Split out `<NavButtons />`

Tested manually, no changes in behavior that I noticed.

### DIFF
--- a/packages/vue-client/package.json
+++ b/packages/vue-client/package.json
@@ -27,7 +27,7 @@
     "chessground": "^8.3.12",
     "pinia": "^2.1.4",
     "socket.io-client": "^4.7.1",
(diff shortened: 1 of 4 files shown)


merowin (approved):
Very nice 👍

PR #253, opened 2024-04-28
Remove Cypress

benjaminpjones:
We don't have E2E tests.  I tried running the `test:e2e:ci` script for kicks, but that just opened the server, but never performed the tests as far as I can tell.  It's best we don't maintain the dead code.

Also, there's a "moderate" security advisory in one of cypress's dependencies.  I don't believe it's affecting us since we don't run the cypress stuff, but I'd like to make the message go away anyhow.


JonKo314 (approved):
Thanks for keeping it tidy!

PR #255, opened 2024-05-09
Add new boards to Baduk variant

merowin:
- Add Benjitos Graph class (from another branch)
- Add a wrapper for this class -> uses above graph class internally, but adds conversions between (number <-> coordinate) such that this wrapper implements Fillable<CoordinateLike, Color>, which we then work with in Baduk variant.
- Add function to Board factory with this return type.
- Refactor variant Baduk s.t. it works with the new boards, by using the fillable interface.
- Refactor variants that inherit Baduk:
  - Drift, Keima, Pyramid can't work with arbitrary boards
    1. Validate in constructor that we work with a grid board config
    2. Use type assertion and work with Grid as before
  - Refactor other variants s.t. they work with the fillable interface

### Still needs to be done:
- adapt the board components in vue-client
- add config input forms for the variants that now can work with arbitrary boards
- test all kinds of stuff!
- improve the config typing: currently config may not have a board property (-> handled as grid config for backwards compatibility), and if it has a board property, then the outer properties width, height are ignored currently

### DIFF
--- a/packages/shared/src/lib/graph.ts
+++ b/packages/shared/src/lib/graph.ts
@@ -84,6 +84,10 @@ export class Graph<T> implements Fillable<number, T> {
       initialValue,
     );
   }
+
+  serialize() {
+    return [...this.data];
+  }
 }
 
 function identity<T>(x: T): T {
@@ -123,7 +127,7 @@ export class GraphWrapper<T> implements Fillable<CoordinateLike, T> {
   set(index: CoordinateLike, val: T): void {
     this.graph.set(this.unpackCoordinate(index), val);
   }
-  map<TDest>(f: (val: T) => TDest): Fillable<CoordinateLike, TDest> {
+  map<TDest>(f: (val: T) => TDest): GraphWrapper<TDest> {
     return new GraphWrapper(this.graph.map(f));
   }
   neighbors(index: CoordinateLike): Coordinate[] {
@@ -138,5 +142,10 @@ export class GraphWrapper<T> implements Fillable<CoordinateLike, T> {
     return this.graph.forEach((value: T, index: number) =>
       f(value, this.packNumber(index)),
     );
+  }
+
+  serialize() {
+    // Return a 2D array, since this looks a lot like a Grid.
+    return [this.graph.serialize()];
   }
 }
--- a/packages/shared/src/lib/graph.ts
+++ b/packages/shared/src/lib/graph.ts
@@ -148,4 +148,20 @@ export class GraphWrapper<T> implements Fillable<CoordinateLike, T> {
     // Return a 2D array, since this looks a lot like a Grid.
     return [this.graph.serialize()];
   }
+
+  reduce<OutT>(
+    callbackfn: (
+      previousValue: OutT,
+      currentValue: T,
+      index: CoordinateLike,
+      array: GraphWrapper<T>,
+    ) => OutT,
+    initialValue: OutT,
+  ): OutT {
+    return this.graph.reduce(
+      (previousValue, currentValue, index) =>
+        callbackfn(previousValue, currentValue, this.packNumber(index), this),
+      initialValue,
+    );
+  }
 }
--- a/packages/shared/src/lib/graph.ts
+++ b/packages/shared/src/lib/graph.ts
@@ -1,0 +1,148 @@
+import { Coordinate, CoordinateLike } from "./coordinate";
+import { Fillable } from "./group_utils";
+
+export type SerializedGraph<T> = { adjacency_array: number[][]; data: T[] };
+
+export class Graph<T> implements Fillable<number, T> {
+  private data: T[];
+  private adjacency_array: number[][];
+  constructor(adjacency_array: number[][]) {
+    this.adjacency_array = copy2D(adjacency_array);
+    this.data = new Array(adjacency_array.length);
+  }
+
+  at(index: number): T | undefined {
+    return this.data.at(index);
+  }
+
+  set(index: number, value: T): void {
+    this.data[index] = value;
+  }
+
+  map<S>(
+    callbackfn: (value: T, index: number, grid: Graph<T>) => S,
+    thisArg?: this,
+  ): Graph<S> {
+    const ret = new Graph<S>(this.adjacency_array);
+    ret.data = this.data.map(
+      (value: T, index: number) => callbackfn(value, index, this),
+      thisArg,
+    );
+    return ret;
+  }
+
+  forEach(
+    callbackfn: (value: T, index: number, grid: Graph<T>) => void,
+    thisArg?: this,
+  ): void {
+    this.data.forEach((value: T, index: number) => {
+      callbackfn(value, index, this);
+    }, thisArg);
+  }
+
+  static fromPlainObject<T>(o: SerializedGraph<T>) {
+    if (o.adjacency_array.length != o.data.length) {
+      throw new Error("Adjacency array must be the same length as data array");
+    }
+    const ret = new Graph<T>(o.adjacency_array);
+    ret.data = copyArray(o.data);
+    return ret;
+  }
+
+  export(): SerializedGraph<T> {
+    return {
+      adjacency_array: copy2D(this.adjacency_array),
+      data: copyArray(this.data),
+    };
+  }
+
+  fill(val: T, start?: number, end?: number): Graph<T> {
+    this.data.fill(val, start, end);
+    return this;
+  }
+
+  neighbors(index: number) {
+    return copyArray(this.adjacency_array[index]);
+  }
+
+  isInBounds(index: number) {
+    return index >= 0 && index < this.data.length;
+  }
+
+  reduce<OutT>(
+    callbackfn: (
+      previousValue: OutT,
+      currentValue: T,
+      index: number,
+      array: Graph<T>,
+    ) => OutT,
+    initialValue: OutT,
+  ): OutT {
+    return this.data.reduce(
+      (previousValue, currentValue, index) =>
+        callbackfn(previousValue, currentValue, index, this),
+      initialValue,
+    );
+  }
+
+  // I believe this method will be useful for rendering the board
+  // and this typing should make sure that this.data is not changed from outside.
+  getDataIterator(): Iterable<T> {
+    return this.data;
+  }
+}
+
+function identity<T>(x: T): T {
+  return x;
+}
+function copyArray<T>(a: T[]): T[] {
+  return a.map(identity);
+}
+function copy2D<T>(a: T[][]): T[][] {
+  return a.map(copyArray);
+}
+
+export class GraphWrapper<T> implements Fillable<CoordinateLike, T> {
+  private graph: Graph<T>;
+
+  constructor(graph: Graph<T>) {
+    this.graph = graph;
+  }
+
+  private packNumber(num: number): Coordinate {
+    return new Coordinate(num, 0);
+  }
+
+  private unpackCoordinate(coord: CoordinateLike): number {
+    if (coord.y !== 0) {
+      throw Error(
+        "Implementation Error: Graph Wrapper accepty only coordinates with y = 0.",
+      );
+    }
+
+    return coord.x;
+  }
+
+  at(index: CoordinateLike): T | undefined {
+    return this.graph.at(this.unpackCoordinate(index));
+  }
+  set(index: CoordinateLike, val: T): void {
+    this.graph.set(this.unpackCoordinate(index), val);
+  }
+  map<TDest>(f: (val: T) => TDest): Fillable<CoordinateLike, TDest> {
+    return new GraphWrapper(this.graph.map(f));
+  }
+  neighbors(index: CoordinateLike): Coordinate[] {
+    return this.graph
+      .neighbors(this.unpackCoordinate(index))
+      .map(this.packNumber);
+  }
+  isInBounds(index: CoordinateLike): boolean {
+    return this.graph.isInBounds(this.unpackCoordinate(index));
+  }
+  forEach(f: (value: T, index: CoordinateLike) => void): void {
+    return this.graph.forEach((value: T, index: number) =>
+      f(value, this.packNumber(index)),
+    );
+  }
+}
--- a/packages/shared/src/lib/graph.ts
+++ b/packages/shared/src/lib/graph.ts
@@ -84,12 +84,6 @@ export class Graph<T> implements Fillable<number, T> {
       initialValue,
     );
   }
-
-  // I believe this method will be useful for rendering the board
-  // and this typing should make sure that this.data is not changed from outside.
-  getDataIterator(): Iterable<T> {
-    return this.data;
-  }
 }
 
 function identity<T>(x: T): T {
--- a/packages/shared/src/variants/baduk.ts
+++ b/packages/shared/src/variants/baduk.ts
@@ -5,6 +5,7 @@ import { SuperKoDetector } from "../lib/ko_detector";
 import { AbstractGame } from "../abstract_game";
 import {
   BoardConfig,
+  BoardPattern,
   createBoard,
   createGraph,
 } from "../lib/abstractBoard/boardFactory";
@@ -249,5 +250,16 @@ export class GridBaduk extends Baduk {
       throw "GridBaduk requires a GridBadukConfig";
     }
     super(config);
+  }
+
+  override defaultConfig(): GridBadukConfig {
+    return {
+      komi: 6.5,
+      board: {
+        type: BoardPattern.Grid,
+        width: 19,
+        height: 19,
+      },
+    };
   }
 }
--- a/packages/shared/src/variants/baduk.ts
+++ b/packages/shared/src/variants/baduk.ts
@@ -1,8 +1,17 @@
 import { Coordinate, CoordinateLike } from "../lib/coordinate";
 import { Grid } from "../lib/grid";
-import { getGroup, getOuterBorder } from "../lib/group_utils";
+import { Fillable, getGroup, getOuterBorder } from "../lib/group_utils";
 import { SuperKoDetector } from "../lib/ko_detector";
 import { AbstractGame } from "../abstract_game";
+import {
+  BoardConfig,
+  BoardPattern,
+  GridBoardConfig,
+  createBoard,
+  createGraph,
+} from "../lib/abstractBoard/boardFactory";
+import { Intersection } from "../lib/abstractBoard/intersection";
+import { GraphWrapper } from "../lib/graph";
 
 export enum Color {
   EMPTY = 0,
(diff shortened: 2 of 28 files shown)


packages/shared/src/variants/baduk.ts
```diff
@@ -43,18 +56,23 @@ export class Baduk extends AbstractGame<BadukConfig, BadukState> {
...
+      );
+    } else {
+      this.boardAsIntersections = createBoard(this.config.board!, Intersection);
+      this.board = new GraphWrapper(createGraph(this.boardAsIntersections));
+    }
   }
 
   override exportState(): BadukState {
     return {
-      board: this.board.to2DArray(),
```
benjaminpjones, line 53: I'd really like to keep the exports "JSON-compatible".  What I mean by that is that information should not be lost by converting to JSON:

```
let grid: Grid<number> = ...;
let arr: number[][] = ...;

grid = JSON.parse(JSON.stringify(grid));  // not actually a Grid
arr = JSON.parse(JSON.stringify(arr)); // a 2D Array, just like the original
```

**Motivation**:  We need to keep in mind that game state needs to travel over the wire at some point (#48, #174), and we need a way to losslessly encode the state.

In my mind, the way to do that is to return simple types from `exportState()` with helper methods like `Grid.prototype.to2DArray()`, then use the JSON utils to encode and decode.  (If you need a `Graph`-compatible way to do this, feel free to alias that method to `Grid.prototype.serialize()`).

And let me know if you'd propose another way to pass state over the wire that doesn't simplification inside exportState!

benjaminpjones:
This is so great by the way, I'm looking forward to playing some more graph Go!  And I really appreciate the effort you put into de-duplicating implementations - this will help maintainability immensely.

benjaminpjones:
One thought to keep this PR small and atomic - if you introduce a GridBaduk [adapter], you won't need to change all the downstream stuff (vue boards, config forms, etc) at the same time.  Something like (I think the types work.. [ts playground](https://www.typescriptlang.org/play/?#code/PTAEHUFMBsGMHsC2lQBd5oBYoCoE8AHSAZVgCcBLA1UABWgEM8BzM+AVwDsATAGiwoBnUENANQAd0gAjQRVSQAUCEmYKsTKGYUAbpGF4OY0BoadYKdJMoL+gzAzIoz3UNEiPOofEVKVqAHSKymAAmkYI7NCuqGqcANag8ABmIjQUXrFOKBJMggBcISGgoAC0oACCbvCwDKgU8JkY7p7ehCTkVDQS2E6gnPCxGcwmZqDSTgzxxWWVoASMFmgYkAAeRJTInN3ymj4d-jSCeNsMq-wuoPaOltigAKoASgAywhK7SbGQZIIz5VWCFzSeCrZagNYbChbHaxUDcCjJZLfSDbExIAgUdxkUBIursJzCFJtXydajBDIKMjJBhLABimMY0ncoAA3ooSiVmJBUAAKVb5frsRDSb78PACzhCkVkACUEql3wA3IoAL7BWCMQTCADilFcbI580oOjqzjIZHlwu+AG0ALp20AAXlA1utAEZ+AAmfgAZlt-GtABZ+ABWfgANn9LoA7PwABz8ACctttysNXN5-MFVrIYst0plrNATlQ+MyakEAUcZGteHtq1ToDVhsE3L5+dFoHF2el-BN0HYkA7sqLQ0r1dr9dtTtA-cHiqbqsUGoYWtAuoYBE0Bo5BGNprE5uHDud7q9vv4wdAYdA4f4sdACdAybTHIz7Z7ne7kpzhZZxe5MsBHHc1rQbBdmw5VtM2HPNP1zWcGAHId4L-YCq1AhsZznSAIKXZdNWEAAhBhuHYRIdxKYFHG4AV6WgRl3FfEoEE4QRUDIdhYHQMgeVQAUACJWAobgBNAAAfUAhLITdMAEv92UNDkEVAPinUdZ1pJE+TWUUpSlLHAJqLIVxnU4SAJHXPUeRlZilJVcFoFbXT9P0wzjNM-oLKs2SbLsw1IJKQKjQpGyXNc1jBHgdwAmgeBmD4isjPgGiAnfD1QDdGVbL0xc1QI1cdT1EiyMSNYFB4YjSPI8KShUbhIGSDJ5EsdoxC1b56kacZIFqdhnKkUB4gGSz5AAcmEQqKGYczXAyK52A2GzcpUTB4CkPQEK+PpECYHqrgYJFsVyPBhCsbgMDHXKPIAQgFXURLsyKOK4ni+ME4TRIU1zQBUBQNE4dQkOgPB+FQNrjnYyBEDEaBJm4PBRlQDR9GA3hcsNFRpHYGhOO2KFnA6sguq8Xb4lRsdwVWDjxEBJEMeU1I1JujSpM+nTKJ+rA2AkHkBIe7gSpq2ovEaEGxFgCxqC0PU0HaQR5IZoKlYWpbUBygLgny57xmqxIzO8gWhfiPn2Zy6Q9YCPdQtsoA)):

```
class GridBaduk extends Baduk {
    board!: Grid<Color>;
    constructor(config: BadukConfig) {
        if (!isGridConfig(config)) {
            throw("GridBaduk only accepts grid configs!");
        }
        super(config);
    }
}
```

This will also help with variants like Keima and Pyramid, where currently you are relying on type assertion at the access sites (`this.board as Grid<Color>`)

packages/shared/src/variants/baduk.ts
```diff
@@ -196,7 +217,10 @@ export class Baduk extends AbstractGame<BadukConfig, BadukState> {
 }
 
 /** Returns true if the group containing (x, y) has at least one liberty. */
-export function groupHasLiberties(group: CoordinateLike[], board: Grid<Color>) {
+export function groupHasLiberties(
+  group: CoordinateLike[],
+  board: Fillable<CoordinateLike, Color>,
```
JonKo314, line 222: Should this be `BadukBoard<Color>` instead of `Fillable<...>`?

packages/shared/src/variants/baduk.ts
```diff
@@ -206,3 +230,16 @@ export function groupHasLiberties(group: CoordinateLike[], board: Grid<Color>) {
 function count_color<T>(value: T) {
   return (total: number, color: T) => total + (color === value ? 1 : 0);
 }
+
+export type GridBadukConfig = {
+  width: number;
+  height: number;
+  komi: number;
+  board?: GridBoardConfig;
+};
```
JonKo314, line 239: Can we use something like
```
export type GridBadukConfig = {
  width: number;
  height: number;
  komi: number;
} | {
  board: GridBoardConfig;
  komi: number;
}
```

otherwise there's room for trouble:
```
{
  komi: 6.5;
  width: 9;
  height: 9;
  board: {
    type: "grid";
    width: 19;
    height: 19;
  }
}
```

packages/shared/src/variants/drift.ts
```diff
@@ -1,6 +1,13 @@
...
-import { Baduk, BadukConfig, Color, groupHasLiberties } from "./baduk";
+import {
+  Baduk,
+  BadukConfig,
+  Color,
+  groupHasLiberties,
+  isGridBadukConfig,
+} from "./baduk";
 
 export interface DriftGoConfig extends BadukConfig {
```
JonKo314, line 12: Can this extend `GridBadukConfig` instead?

packages/shared/src/variants/drift.ts
```diff
@@ -11,6 +18,9 @@ export class DriftGo extends Baduk {
   private typedConfig: DriftGoConfig;
 
   constructor(config?: DriftGoConfig) {
+    if (config && !isGridBadukConfig(config)) {
+      throw Error("config for drift must be for board type grid");
+    }
     super(config);
     this.typedConfig = config ?? this.defaultConfig();
   }
```
JonKo314, line 26: I think we can change the type of the board to a Grid here.

See
```
class Board {
  foo() {
    return "graph";
  }
}

class SpecificBoard extends Board {
  foo() { return "grid"; }
  bar() { return "asdf"; }
}

class GameParent {
  constructor(protected board: Board) {

  }

  public foo() {
    return this.board.foo();
  }
}

class GameChild extends GameParent {
  constructor(protected board: SpecificBoard) {
    super(board);
  }

  public bar() {
    return this.board.bar();
  }
}

const game = new GameChild(new SpecificBoard());
console.log(game.foo()); // prints "grid"
console.log(game.bar()); // prints "asdf"
```

Then we would need the `as Grid` type assertions.

packages/shared/src/variants/quantum.ts
```diff
@@ -23,7 +23,7 @@ class BadukHelper {
     this.badukGame.playMove(player, "pass");
   }
 
-  play(player: number, move: string): Coordinate[] {
```
benjaminpjones, line 26: Nit: I like the rule ["Return the most specific type, accept the most generic type"](https://enterprisecraftsmanship.com/posts/return-the-most-specific-type/).  In other words, I usually accept `CoordinateLike` as argument, and return `Coordinate`.  That way, callers have access to all the utility methods of the `Coordinate` class, but only need to pass the bare minimum (`{x: 1, y: 2}`) as a parameter.

packages/shared/src/variants/baduk.ts
```diff
@@ -14,23 +23,27 @@ export interface BadukConfig {
...
   next_to_play: 0 | 1;
   captures: { 0: number; 1: number };
   last_move: string;
-  score_board?: Color[][];
+  score_board?: BadukBoardType<Color>;
 }
 
 export type BadukMove = { 0: string } | { 1: string };
 
+export declare type BadukBoardType<TColor> = Fillable<CoordinateLike, TColor>;
```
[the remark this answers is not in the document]
merowin: renamed

packages/shared/src/variants/baduk.ts
```diff
@@ -170,12 +188,15 @@ export class Baduk extends AbstractGame<BadukConfig, BadukState> {
...
+    board.forEach(
+      (intersection) => (black_points += intersection === Color.BLACK ? 1 : 0),
+    );
+
+    let white_points = 0;
+    board.forEach(
+      (intersection) => (white_points += intersection === Color.WHITE ? 1 : 0),
     );
-    const white_points: number =
-      board.reduce(count_color<Color>(Color.WHITE), 0) + this.config.komi;
```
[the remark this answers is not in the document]
merowin: oops, thank you for catching that! fixed

packages/shared/src/variants/baduk.ts
merowin: changed type nominally

packages/shared/src/variants/baduk.ts
merowin: done

packages/shared/src/variants/drift.ts
merowin: changed to type and extend GridBadukConfig

packages/shared/src/variants/quantum.ts
```diff
@@ -211,24 +211,20 @@ export class QuantumGo extends AbstractGame<BadukConfig, QuantumGoState> {
...
-        prevBoard.at({ x, y }) !== Color.EMPTY &&
-        currBoard.at({ x, y }) === Color.EMPTY
-      ) {
-        captures.push(new Coordinate(x, y));
-      }
+  prevBoard: BadukBoardType<Color>,
+  currBoard: BadukBoardType<Color>,
+): CoordinateLike[] {
+  const captures: CoordinateLike[] = [];
+  prevBoard.forEach((color, coordinate) => {
```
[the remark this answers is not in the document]
merowin: Thank you

packages/shared/src/variants/baduk.ts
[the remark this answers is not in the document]
merowin: renamed to BadukBoard

packages/shared/src/variants/baduk.ts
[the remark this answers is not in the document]
merowin: Oops, thank you for catching that! Fixed

packages/shared/src/variants/baduk.ts
merowin: changed type

packages/shared/src/variants/tetris.ts
```diff
@@ -12,7 +11,7 @@ export class TetrisGo extends Baduk {
   }
 }
 
-function getGroup(pos: CoordinateLike, board: Grid<Color>) {
+function getGroup(pos: CoordinateLike, board: BadukBoardType<Color>) {
```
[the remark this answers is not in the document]
merowin: Right, thanks for pointing this out! Replaced by group_utils.getGroup()

packages/shared/src/variants/quantum.ts
merowin: Changed the types in quantum.ts thusly

merowin:
Ok, I think I'm at a point where we can (hopefully) merge this. Here's an overview of what I did since yesterday:

- Merge in Bens Baduk adapter class
- change type of Baduk config such that it uses the same logic as the board factory expects, or has the old type
-> The goal is to get rid of the legacy type with a DB migration in future, and support both in the meantime
- Fix the introduced type errors with a function that extracts width, height from a GridBadukConfig
- The config forms already emit a config for the new type structure, so this will impact the structure of newly created games in the DB.
- Try to adress the code review remarks
- Test that the site doesn't crash

packages/shared/src/variants/baduk.ts
```diff
@@ -170,12 +188,15 @@ export class Baduk extends AbstractGame<BadukConfig, BadukState> {
...
-      0,
+    let black_points = 0;
+    board.forEach(
+      (intersection) => (black_points += intersection === Color.BLACK ? 1 : 0),
+    );
+
+    let white_points = 0;
+    board.forEach(
+      (intersection) => (white_points += intersection === Color.WHITE ? 1 : 0),
     );
```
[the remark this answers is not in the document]
JonKo314: Alternatively, we could also have `Fillable` return a one-dimensional array and use `reduce` on that.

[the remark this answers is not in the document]
merowin: I'm fine with any of these options, so I'd like to leave the decision to you.

We can also undo the changes here and continue using this type instead of fillable:

`export declare type BadukBoard<TColor> = Grid<TColor> | GraphWrapper<TColor>;`

packages/shared/src/lib/graph.ts
```diff
@@ -0,0 +1,142 @@
...
+  return x;
+}
+function copyArray<T>(a: T[]): T[] {
+  return a.map(identity);
+}
+function copy2D<T>(a: T[][]): T[][] {
+  return a.map(copyArray);
+}
+
+export class GraphWrapper<T> implements Fillable<CoordinateLike, T> {
```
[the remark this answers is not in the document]
merowin: What do you think about introducing a new interface that extends fillable? It could be our board interface and have methods like serialize, reduce etc. that maybe don't fit fillable, but make sense for our board interface?
Then this interface could also handle the encoding / decoding of moves.

packages/shared/src/variants/baduk.ts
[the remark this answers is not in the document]
benjaminpjones: Oh sure if that allows us to use `reduce` again, typing as `Grid | GraphWrapper` would be nice!

In the end I don't have a strong preference - I do want to keep `reduce` so we can re-use the reducer and keep variables like `white_points` `const`, but otherwise I don't feel strongly about how we do it.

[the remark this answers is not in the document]
merowin: done!

packages/shared/src/variants/baduk.ts
```diff
@@ -190,19 +216,48 @@ export class Baduk extends AbstractGame<BadukConfig, BadukState> {
...
+  board: BadukBoard<Color>,
+) {
   const outer_border = getOuterBorder(group, board);
   const border_colors = outer_border.map((pos) => board.at(pos));
   return border_colors.includes(Color.EMPTY);
 }
 
 /** Returns a reducer that will count occurences of a given number **/
+// This was used before... will need to figure out where it went
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
benjaminpjones, line 236:
```suggestion
```

I added these comments so suppress the linter.  Should be good to remove now that we reverted the `reduce(count_color())` code

packages/shared/src/variants/baduk.ts
```diff
@@ -190,19 +216,48 @@ export class Baduk extends AbstractGame<BadukConfig, BadukState> {
     this.phase = "gameover";
   }
 
-  defaultConfig(): BadukConfig {
+  defaultConfig(): GridBadukConfig {
     return { width: 19, height: 19, komi: 6.5 };
```
benjaminpjones, line 220: If top level `width` and `height` are now considered legacy, we may want to update this return value.

packages/shared/src/variants/quantum.ts
```diff
@@ -32,18 +33,18 @@ class BadukHelper {
...
-  clear(group: Coordinate[]) {
+  clear(group: CoordinateLike[]) {
     for (const pos of group) {
       this.badukGame.board.set(pos, Color.EMPTY);
     }
   }
 }
 
-export class QuantumGo extends AbstractGame<BadukConfig, QuantumGoState> {
+export class QuantumGo extends AbstractGame<GridBadukConfig, QuantumGoState> {
```
benjaminpjones, line 43: Doesn't have to change in this PR, but I think QuantumGo could be adapted to graph boards.

packages/vue-client/src/components/boards/MulticolorGridBoard.vue
```diff
@@ -15,7 +15,7 @@ import { positionsGetter } from "./board_utils";
 const props = defineProps<{
   board: (MulticolorStone | null)[][];
   background_color?: string;
-  config: { height: number; width: number };
+  config: { width: number; height: number };
```
benjaminpjones, line 18: Perhaps change this property name to dimensions since it's no longer representative of the game config.

benjaminpjones (approved):
Hey looks good, no blocking comments.  I haven't tested the UI changes, but I trust you will at least run through the game creation flow with one of the affected variants.

Thanks for doing this!

packages/shared/src/variants/baduk.ts
merowin: removed comments

packages/shared/src/variants/baduk.ts
merowin: changed to return new type structure

packages/shared/src/variants/quantum.ts
merowin: That's the plan eventually! I believe this will work as soon as I changed the vue-client components for rendering the board for Baduk. But that's for another PR.

packages/vue-client/src/components/boards/MulticolorGridBoard.vue
merowin: changed props name to board_dimensions

PR #256, opened 2024-05-10
Revert to 2D array board and introduce GridBaduk class

benjaminpjones:
@merowin, this fixes the build errors by introducing a `GridBaduk` adapter class.  This class has the same interface as the original `Baduk` class.


merowin (approved):
Thank you, I got carried away and tried to do too many things at once.

PR #259, opened 2024-05-13
add board config input form components

merowin:
This adds components for input forms for the various board configs. I tried to streamline these components by having one dedicated component for each board pattern.
With this it will be possible to create and play games with the new board patterns for variants baduk, phantom, capture, tetris, thue-morse, and one-color.
For variants freeze and quantum it should not be too much work to also enable this, but since these have their own board component, I'd like to do this in a separate PR.

It's worth noting that there's a couple of possible improvements that I've ignored for now.
1.) The graph board component doesn't use the annotations yet.
2.) At the moment, the newly added config input forms don't use the default config as defined by the variant itself, but their own default config.


benjaminpjones:
> I tried to streamline these components by having one dedicated component for each board pattern.

Solid design choice.

> With this it will be possible to create and play games with the new board patterns for variants baduk, phantom, capture, tetris, thue-morse, and one-color.

Nice to see it all come together!

Issue #151, opened 2023-09-12
mosaigo


Ruyaoo:
![百度贴吧(1)](https://github.com/govariantsteam/govariants/assets/102702890/71d715ed-d5eb-4002-a9d0-001968c940bf)
![百度贴吧(2)](https://github.com/govariantsteam/govariants/assets/102702890/9f114477-60d1-41fa-ac03-0003a97e3110)
![百度贴吧(3)](https://github.com/govariantsteam/govariants/assets/102702890/66e080a3-126a-4c44-87ce-5b252525b53c)
![百度贴吧(4)](https://github.com/govariantsteam/govariants/assets/102702890/669e7b2d-8aef-4f07-9b9d-ca35a1686355)

Hello,May I ask if you are interested in mosaigo.It can be increased or eliminated the line of board.

JonKo314:
Hello, I'm definitely interested in variants with a board like the second picture (the wooden board) and I think it shouldn't be that hard to achieve with our current code.

What is the meaning of the dark squares on the other boards? Do they also represent a border?

Ruyaoo:
Yes, one way is to remove the lines.

One way is to remove blocks.

JonKo314:
I'm not sure what you mean by removing blocks. When we look at a Goban as a [graph](https://en.wikipedia.org/wiki/Graph_(discrete_mathematics)), all we can do is (add or) remove nodes and edges.

Ruyaoo:
yes,thank you,good job!



---Original---
From: ***@***.***&gt;
Date: Sun, Nov 12, 2023 06:38 AM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [govariantsteam/govariants] mosaigo (Issue #151)




 
I'm not sure what you mean by removing blocks. When we look at a Goban as a graph, all we can do is (add or) remove nodes and edges.

benjaminpjones:
> I'm not sure what you mean by removing blocks. When we look at a Goban as a [graph](https://en.wikipedia.org/wiki/Graph_(discrete_mathematics)), all we can do is (add or) remove nodes and edges.

I believe the method of removing "blocks" is as follows:

1) A square surrounded by 4 edges is called a "block".  For example, a 9x9 board is made up of 8x8=64 blocks.
2) Blocks may be removed for mosaigo.
3) In terms of graphs, an edge touches one or two blocks.  If all adjacent blocks are removed, the edge is also removed.
[...]

PR #262, opened 2024-05-16
Grid with holes

JonKo314:
Adds new graph board type "grid with holes", that allows for missing intersections or missing connections between intersections. Tested for Baduk. See #151.

### DIFF
--- a/packages/shared/src/lib/abstractBoard/boardFactory.ts
+++ b/packages/shared/src/lib/abstractBoard/boardFactory.ts
@@ -6,6 +6,7 @@ import { CreateCircularBoard } from "../../variants/badukWithAbstractBoard/abstr
 import { TrihexagonalBoardHelper } from "../../variants/badukWithAbstractBoard/abstractBoard/TrihexagonalBoardHelper";
 import { createSierpinskyBoard as createSierpinskyBoardExternal } from "../../variants/badukWithAbstractBoard/abstractBoard/SierpinskyBoard";
 import { Graph } from "../graph";
+import { Grid } from "../grid";
 
 export const BoardPattern = {
   Grid: "grid",
@@ -13,6 +14,7 @@ export const BoardPattern = {
   Circular: "circular",
   Trihexagonal: "trihexagonal",
   Sierpinsky: "sierpinsky",
+  GridWithHoles: "gridWithHoles",
 } as const;
 
 export type BoardConfig =
@@ -20,7 +22,8 @@ export type BoardConfig =
   | RhombitrihexagonalBoardConfig
   | CircularBoardConfig
   | TrihexagonalBoardConfig
-  | SierpinskyBoardConfig;
+  | SierpinskyBoardConfig
+  | GridWithHolesBoardConfig;
 
 export interface GridBoardConfig {
   type: typeof BoardPattern.Grid;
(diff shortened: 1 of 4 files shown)


packages/vue-client/src/components/boards/MulticolorGraphBoard.vue
```diff
@@ -21,7 +21,9 @@ const props = defineProps<{
   board_config: BoardConfig;
 }>();
 
-const intersections = createBoard(props.board_config, Intersection);
+const intersections = computed(() =>
```
merowin, line 24: Putting the intersections into a computed caused performance problems for me. For the game page it should be enough to compute the intersections once.
For the variant demo I noticed that this causes a problem, but it would be great if we can find another solution that doesn't slow the game page down.

merowin:
Looks great, can't wait to try it!

packages/shared/src/lib/abstractBoard/boardFactory.ts
```diff
@@ -49,6 +52,11 @@ export interface SierpinskyBoardConfig {
   size: number;
 }
 
+export interface GridWithHolesBoardConfig {
+  type: "gridWithHoles";
+  bitmap: (0 | 1 | 2 | 3 | 4)[][];
```
benjaminpjones, line 57: I'm not sure I understand the bitmap. What do the different values represent?

merowin: Might be explained by this forum post (and the two after): https://forums.online-go.com/t/collective-development-of-a-server-for-variants/43682/145

packages/vue-client/src/components/boards/MulticolorGraphBoard.vue
benjaminpjones: `computed` (or some sort of reactivity) seems like an appropriate thing to do here.  What kind of performance issues were you seeing?

> For the game page I think it would be enough to compute the intersections once.

I'm not exactly sure how Vue router works, but if we ever support navigating to different games from the games page, this may lead to stale data.

merowin: I remember noticing a lag between clicking (for playing a move), and the stone appearing. But I just tried to reproduce, and didn't have such problems. So now I'm thinking this is probably fine.

But I think its worth noting that: Whenever a move is made, the server sends all game information to the clients. The config hasn't changed (but the client doesn't know that) and so the intersections are computed anew. I believe this also causes the board to be rerendered, since it depends on the intersections.

packages/shared/src/lib/abstractBoard/boardFactory.ts
JonKo314: I've added a comment to explain the values.

packages/vue-client/src/components/GameCreation/BoardConfigForms/GridWithHolesBoardConfigForm.vue
```diff
@@ -0,0 +1,82 @@
...
+const config = ref(props.initialConfig ?? defaultConfig);
+const bitmapString = ref<string>(JSON.stringify(defaultConfig.bitmap));
+const configError = ref<string>();
+
+watch(bitmapString, () => {
+  console.log(`bitmapString changed: ${bitmapString.value}`);
+  configError.value = undefined;
+  try {
+    const j = JSON.parse(bitmapString.value);
+    if (!Array.isArray(j) || j.some((row) => !Array.isArray(row))) {
```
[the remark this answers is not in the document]
JonKo314: Done. Thanks for the code, almost perfect.

benjaminpjones:
Any plans to merge this one?

PR #267, opened 2024-06-02
send state via api instead of all moves

merowin:
### Changes
- Add server endpoints for getting (initial-) game information. Importantly this does not include the moves array, instead the view state defining properties. So the received information can depend on the selected seat.
- add a subscription based system for sending the state to the appropriate clients when a move is made. I believe this does not fully work yet, unfortunately.
- major change of the game view page to work with the new structures

### Remarks:
- I did not remove the endpoint that returns the game moves array. I'm not sure if its used by the game list. We may need to change that too, to make sure that hidden information can't leak.
- I tested (among other things) the staged moves of parallel. At the moment they don't show up in the game view unfortunately, I don't know the reason yet.
- In the game view I've added a map that stores all known values of states depending on (round|seat)-combinations. My idea is to avoid making redundant network calls with this. But I'm not sure if it does what I want it to do.

### DIFF
--- a/packages/server/src/games.ts
+++ b/packages/server/src/games.ts
@@ -6,6 +6,7 @@ import {
   getOnlyMove,
   HasTimeControlConfig,
   GamesFilter,
+  AbstractGame,
 } from "@ogfcommunity/variants-shared";
 import { ObjectId, WithId, Document, Filter } from "mongodb";
 import { getDb } from "./db";
@@ -17,6 +18,7 @@ import {
 } from "./time-control/time-control";
 import { timeControlHandlerMap } from "./time-control/time-handler-map";
 import { Clock } from "./time-control/clock";
+import { GameStateResponse } from "../../shared/src/api_types";
 
 export function gamesCollection() {
   return getDb().db().collection("games");
@@ -193,9 +195,41 @@ export async function handleMoveAndTime(
 
   game.moves.push(moves);
 
-  io().emit(`game/${game_id}`, game);
+  emitGame(game.id, game.players?.length ?? 0, game_obj);
 
   return game;
+}
+
+function emitGame(
+  game_id: string,
+  num_players: number,
+  game_obj: AbstractGame<unknown, unknown>,
+): void {
+  const next_to_play = game_obj.nextToPlay();
+  const specialMoves = game_obj.specialMoves();
+
+  io()
+    .to(`game/${game_id}/`)
+    .emit("move", {
+      state: game_obj.exportState(null),
+      round: game_obj.round,
+      next_to_play: next_to_play,
+      special_moves: specialMoves,
+      result: game_obj.result,
+      seat: null,
+    });
+
+  for (let seat = 0; seat < num_players; seat++) {
+    const gameStateResponse: GameStateResponse = {
+      state: game_obj.exportState(seat),
+      round: game_obj.round,
+      next_to_play: next_to_play,
+      special_moves: specialMoves,
+      result: game_obj.result,
+      seat: seat,
+    };
+    io().to(`game/${game_id}/${seat}`).emit("move", gameStateResponse);
+  }
 }
 
 async function updateSeat(
@@ -237,6 +271,34 @@ export async function leaveSeat(
   user_id: string,
 ) {
   return updateSeat(game_id, seat, user_id, undefined);
+}
+
+export async function getGameState(
+  game: GameResponse,
+  seat: number | null,
+  round: number | null,
+): Promise<GameStateResponse> {
+  const game_obj = makeGameObject(game.variant, game.config);
+
+  for (let i = 0; i < game.moves.length; i++) {
+    if (game_obj.round === round) {
+      break;
+    }
+
+    const encoded_move = game.moves[i];
+    const { player, move } = getOnlyMove(encoded_move);
+    game_obj.playMove(player, move);
+  }
+
+  return {
+    state: game_obj.exportState(seat),
+    round: game_obj.round,
+    next_to_play: game_obj.nextToPlay(),
+    special_moves: game_obj.specialMoves(),
+    result: game_obj.result,
+    seat: seat,
+    timeControl: game.time_control,
+  };
 }
 
 // This function exists for games whose players field has not yet been defined.
--- a/packages/server/src/games.ts
+++ b/packages/server/src/games.ts
@@ -209,7 +209,7 @@ function emitGame(
   const specialMoves = game_obj.specialMoves();
 
   io()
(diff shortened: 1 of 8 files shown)


merowin:
Ok, I think I fixed the problems with the state map, as well as the subscriptions.

I've tested and confirmed that navigating through the moves only triggers an API call when we don't have this state yet. So in this respect, the state map works.
I've also tested playing a game with two browsers, and confirmed that we receive the state when a move is played, and that we don't receive information that we shouldn't have.

@benjaminpjones @JonKo314 I'm requesting your review, because this is a central and important feature, and I'm not sure about some choices that I made along the way.

benjaminpjones:
Hey, thanks so much for doing this!

I took a look at the server changes and these seem like reasonable additions to the API.

The Vue stuff is a little more complicated, I'll need to come back to it. Caching state as youve done generally sounds like a good thing though :)

packages/vue-client/src/views/GameView.vue
```diff
@@ -146,40 +146,39 @@ function makeMove(move_str: string) {
...
     return;
   }
-  const message = `game/${props.gameId}`;
-  socket.on(message, (data) => {
-    Object.assign(gameResponse, data);
-  });
+  const topic = `game/${props.gameId}`;
+  socket.emit("subscribe", [topic]);
+  socket.on("move", (state: GameStateResponse) =>
+    setViewState(state.round, state.seat),
```
benjaminpjones, line 159: Shouldn't we call `setNewState` directly?  `setViewState` will trigger another API call, but that's not necessary since we already have a `GameStateResponse`.

benjaminpjones (commented):
I got a chance to test this out, I think there is a regression in parallel go.  When you place a move in parallel, you are supposed to be able to see your move, but with these changes, you get no feedback that a move was played (unless it's the end of the round)


Here is the behavior currently in prod:

https://github.com/govariantsteam/govariants/assets/25233703/d17a6c03-b4c4-4907-b37a-402780f28b18

packages/server/src/games.ts
```diff
@@ -193,11 +195,43 @@ export async function handleMoveAndTime(
...
+  for (let seat = 0; seat < num_players; seat++) {
+    const gameStateResponse: GameStateResponse = {
+      state: game_obj.exportState(seat),
+      round: game_obj.round,
+      next_to_play: next_to_play,
+      special_moves: specialMoves,
+      result: game_obj.result,
+      seat: seat,
+    };
+    io().to(`game/${game_id}/${seat}`).emit("move", gameStateResponse);
```
benjaminpjones, line 231: Does this get used in the client?  Seems like it should, but I didn't see any `emit("subscribe", [`game/${game_id}/${seat}`])`

packages/vue-client/src/views/GameView.vue
merowin: You are right, thank you for catching this!

packages/server/src/games.ts
merowin: Yes the subscription happens in the function setPlayingAs in GameView.vue

benjaminpjones: Totally missed that! Thanks!

merowin:
> I got a chance to test this out, I think there is a regression in parallel go. When you place a move in parallel, you are supposed to be able to see your move, but with these changes, you get no feedback that a move was played (unless it's the end of the round)
> 
> Here is the behavior currently in prod:
> Screen.Recording.2024-06-09.at.12.29.30.AM.mov

Right, thank you for testing and finding this! I believe the problem comes from the getGameState function on the server. It plays moves and breaks if we find that we reached the given round. Then the staged moves are nor included, which is what we want (I believe) for rounds in the past, however not for the current round. This is what makes it a bit tricky to fix.
I tried to fix it using the structuredCopy function (similar as it was used in the vue-client), but it seems that this function is not available on the server - I get an error "structuredClone is undefined".
I'm unsure how to fix it currently.

merowin:
Converting this to draft because I just noticed that I need to remove the circular references in FractionalGoState before merging this!

benjaminpjones:
>it seems that this function is not available on the server - I get an error "structuredClone is undefined". I'm unsure how to fix it currently.

`structuredClone` was added in node 17. I just checked, heroku should be on 18+, but our Github workflows are on 16 😬 Shouldn't be too hard to upgrade though.

If you upgrade your local node version. Does it work?

https://nodejs.org/api/globals.html#structuredclonevalue-options

merowin:
> If you upgrade your local node version. Does it work?

I upgraded Node.js to v. 20.14.0, but structuredClone is still not available. I'll think about how to fix it anyway.

merowin:
I believe that I fixed the problem. I used a naive solution for the gameState endpoint on the server, that is probably not as efficient as possible, but works for now. I've also made it so the state of fractional go is serializable, so the variant can work with this new structure.

There are still many places where users can see the moves (e.g. the server endpoint for making a move returns all moves etc.), so we will need to clean this up as well. But I'd like to do it in a separate pull request.

packages/vue-client/src/views/GameView.vue
```diff
@@ -23,112 +21,114 @@ import PlayersToMove from "@/components/GameView/PlayersToMove.vue";
...
-  });
-  const result =
-    game_obj.phase === "gameover" ? game_obj.result || "Game over" : null;
+function encodeSeatAndRound(round: number, seat: number | null): string {
+  return `${round};${seat ?? ""}`;
+}
 
-  if (state === null) {
-    state = game_obj.exportState(playing_as.value);
+async function setViewState(round: number, seat: number | null): Promise<void> {
```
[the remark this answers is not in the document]
merowin: renamed to updateGameState

packages/vue-client/src/views/GameView.vue
```diff
@@ -23,112 +21,114 @@ import PlayersToMove from "@/components/GameView/PlayersToMove.vue";
...
 
-const DEFAULT_GAME: GameResponse = {
-  id: "",
-  variant: "",
-  moves: [],
-  config: {},
-};
-
+const state_map = new Map<string, GameStateResponse>();
+const view_state = ref<GameStateResponse | undefined>();
```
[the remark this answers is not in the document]
merowin: renamed to game_state.

Thank you!

benjaminpjones (approved):

merowin:
Thank you for your observations! I've created new issues. I'm planning to test a little bit again and make sure that no variant crashes with this. Do you think this is ready for being merged then?

benjaminpjones:
> Do you think this is ready for being merged then?

Yes! :)

PR #272, opened 2024-06-11
Update README.md

benjaminpjones:
This removes/updates some outdated information.

I decide to link to "chess" as the example PR.  It seemed like a good standalone example that uses Vue, although we should create another updated PR at some point (some directories/interfaces have changed since then)


benjaminpjones:
cc:@SameerDalal sorry for the delay, this should be a lot more up-to-date

PR #273, opened 2024-06-11
inital sgf download

SameerDalal:
Currently the SGF data for a quantum game is stored in the `this.sgfContent` variable in `quantum.ts`. For the user to download this, I also created a download button in `DownloadSGF.vue`, in the GameView folder, however I'm not sure how to move the contents of the `this.sgfContent` string from the typescript file to the vue file.

### DIFF
--- a/packages/server/src/api.ts
+++ b/packages/server/src/api.ts
@@ -30,9 +30,6 @@ export const router = express.Router();
 // Set up express routes
 
 router.get('/game/:gameId/sgf', async (req, res) => {
-  //route not being called
-  console.log("test");
-  res.status(200).send("test");
   try {
 
     const game = await getGame(req.params.gameId);
@@ -46,10 +43,17 @@ router.get('/game/:gameId/sgf', async (req, res) => {
 
     if (game_obj.getSGF() === '') {
       throw new Error(`SGF not supported for variant ${game.variant}`);
+    } else if (!(game_obj.phase === 'gameover')){
+      throw new Error('Game is not over!');
+    } else {
+      
+      res.set({
+        "Content-Disposition": `attachment; filename="game_${req.params.gameId}.sgf"`,
+      });
+      res.set("Content-Type", "text/plain");
+      res.send(game_obj.getSGF());
     }
-    
-    // download sgf
-    res.download(game_obj.getSGF());
+
   } catch (e) {
     res.status(500).json({ error: e.message });
   }
--- a/packages/server/src/api.ts
+++ b/packages/server/src/api.ts
@@ -43,6 +43,8 @@ router.get('/game/:gameId/sgf', async (req, res) => {
 
     if (game_obj.getSGF() === '') {
       throw new Error(`SGF not supported for variant ${game.variant}`);
+    } else if (game_obj.getSGF() === "non-rectangular"){
+      throw new Error('SGF for non rectangular boards is not possible');
     } else if (!(game_obj.phase === 'gameover')){
       throw new Error('Game is not over!');
     } else {
--- a/packages/server/src/api.ts
+++ b/packages/server/src/api.ts
@@ -29,7 +29,7 @@ export const router = express.Router();
 
 // Set up express routes
 
-router.post('/game/:gameId/sgf', async (req, res) => {
+router.get('/game/:gameId/sgf', async (req, res) => {
   //route not being called
   console.log("test");
   res.status(200).send("test");
@@ -48,8 +48,8 @@ router.post('/game/:gameId/sgf', async (req, res) => {
       throw new Error(`SGF not supported for variant ${game.variant}`);
     }
     
-    // send the SGF string as the response
-    res.status(200).send(game_obj.getSGF());
+    // download sgf
+    res.download(game_obj.getSGF());
   } catch (e) {
     res.status(500).json({ error: e.message });
   }
--- a/packages/server/src/api.ts
+++ b/packages/server/src/api.ts
@@ -8,6 +8,7 @@ import {
   playMove,
   takeSeat,
   leaveSeat,
+  getGameState,
 } from "./games";
 import {
   checkUsername,
@@ -22,6 +23,7 @@ import {
   MovesType,
   User,
   UserResponse,
+  GameInitialResponse,
 } from "@ogfcommunity/variants-shared";
 import { io } from "./socket_io";
 
@@ -29,9 +31,8 @@ export const router = express.Router();
 
 // Set up express routes
 
-router.get('/game/:gameId/sgf', async (req, res) => {
+router.get("/game/:gameId/sgf", async (req, res) => {
   try {
-
     const game = await getGame(req.params.gameId);
 
     const game_obj = makeGameObject(game.variant, game.config);
@@ -41,21 +42,19 @@ router.get('/game/:gameId/sgf', async (req, res) => {
       game_obj.playMove(player, move);
     });
 
-    if (game_obj.getSGF() === '') {
+    if (game_obj.getSGF() === "") {
       throw new Error(`SGF not supported for variant ${game.variant}`);
-    } else if (game_obj.getSGF() === "non-rectangular"){
-      throw new Error('SGF for non rectangular boards is not possible');
-    } else if (!(game_obj.phase === 'gameover')){
-      throw new Error('Game is not over!');
+    } else if (game_obj.getSGF() === "non-rectangular") {
+      throw new Error("SGF for non rectangular boards is not possible");
+    } else if (!(game_obj.phase === "gameover")) {
+      throw new Error("Game is not over!");
     } else {
-      
       res.set({
         "Content-Disposition": `attachment; filename="game_${req.params.gameId}.sgf"`,
       });
       res.set("Content-Type", "text/plain");
       res.send(game_obj.getSGF());
     }
-
   } catch (e) {
     res.status(500).json({ error: e.message });
   }
@@ -84,8 +83,6 @@ router.get("/games", async (req, res) => {
   );
   res.send(games || 0);
 });
-
-
 
 router.post("/games", async (req, res) => {
   const data = req.body;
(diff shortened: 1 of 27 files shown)


benjaminpjones:
This is a nice start!

I think the main part that is missing is the server endpoint.  That will probably go in [server/src/api.ts](https://github.com/govariantsteam/govariants/blob/main/packages/server/src/api.ts).   `games/:gameId` is the most similar endpoint to what we'll need for SGF, except instead of a JSON response, we'll send a text file.

```
router.get('/games/:gameId/sgf', function(request, response){
  try {
    // get the moves and config from database
    const game: GameResponse = await getGame(req.params.gameId);
    // convert to sgf format
    const sgfString: string = toSGFString(game.moves, game.config);

    // pipes a file from memory to the client
    // mostly copied from
    // https://stackoverflow.com/questions/45922074/node-express-js-download-file-from-memory-filename-must-be-a-string
    var fileContents = Buffer.from(sgfString, "base64");
    var fileName = `${req.params.gameId}.sgf`
  
    var readStream = new stream.PassThrough();
    readStream.end(fileContents);

    response.set('Content-disposition', 'attachment; filename=' + fileName);
    response.set('Content-Type', 'text/plain');

    readStream.pipe(response);
  } catch (e) {
    // Express 4 doesn't handle async errors very well unfortunately, so we have to send the error manually
    res.status(500);
    res.json(e.message);
  }
});
```

Disclaimer: this is just me copying code around - I haven't tested myself.  Let me know if you run into any issues with this!

---

I haven't referenced QuantumGo directly.  This is because I noticed your current implementation could work reasonably well for other variants, so it would be a lot simpler to have the SGF implementation decoupled from the variant itself.  If you foresee the implementation diverging from standard SGF/Go, we can figure out a way to let variants define their own SGF function, just let me know.

packages/vue-client/src/components/GameView/DownloadSGF.vue
```diff
@@ -0,0 +1,26 @@
+
+  
+  <script setup lang="ts">
+  import { defineProps } from 'vue';
+  
+  const props = defineProps<{
```
benjaminpjones, line 6: There is a TS error in CI:

```
error TS2345: Argument of type '{}' is not assignable to parameter of type 'NonNullable<(Partial<{ readonly [x: number]: string; }> & Omit<(readonly string[] | { readonly [x: string]: Prop<unknown, unknown> | null | undefined; }) & (VNodeProps & AllowedComponentProps & (ComponentCustomProps & Readonly<...>)), never> & Record<...>) | (Partial<...> & ... 1 more ... & Record<...>)> & Record<...>'.
```

It's a bit vague, but it will be resolved once you add some non-empty props here. (you could also fix the error by removing `defineProps()`, but I think we'll want to pass in `gameId` at least)

merowin:
Nice start, thank you for working on this!

Recently we made some changes that enable graph boards for certain variants, including quantum, see e.g. #264
For grid boards, the moves are stored represented as sgf coordinates, but not for non-rectangular boards.

To distinguish between the two cases, there is the function isGridBadukConfig

SameerDalal:
> > I haven't referenced QuantumGo directly. This is because I noticed your current implementation could work reasonably well for other variants (except maybe the SGF header), so it would be a lot simpler to have the SGF implementation decoupled from the variant itself. If you foresee the implementation diverging from standard SGF/Go, we can figure out a way to let variants define their own SGF function, just let me know.
> 
> On second thought, we probably do need a per-variant implementation, esp. since some variants don't just have black and white. So the way we would do that I think is to expose the SGF via a new optional member of AbstractGame:
> 
> ```
> class AbstractGame {
>     ...
>   sgfContent?: string;
> }
> ```
> 
> Then use that in the express endpoint:
> 
> ```
>   // get the moves and other game info from database
>   const game: GameResponse = await getGame(req.params.gameId);
>   // When game.variant == "quantum", this creates an instance of QuantumGo
>   const game_obj: AbstractGame = makeGameObject(game.variant, game.config);
>   // play out the game
>   game.moves.forEach((moves) => {
>     const { player, move } = getOnlyMove(moves);
>     game_obj.playMove(player, move);
>   });
>   // For backwards compatibility, variants don't have to define sgfContent
>   if (game_obj.sgfContent === undefined) {
>     throw new Error(`SGF not supported for variant ${game.variant}`)
>   }
>   const sgfString: string = game_obj.sgfContent;
> ```

I used this implementation and created the proper API endpoints in api.ts and index.ts. Whenever the download button is clicked the link is changes to `/game/{gameID}/sgf`, which is correct. However in api.ts, the router.get function, which executes the download, is not being called. I also noticed that the other routes have the path `/games` instead of `/game`. I assumed that the `<a>` tag would atuomatically create a GET request to the router, so I'm not sure why the code below is not being called.

api.ts
```
router.get('/game/:gameId/sgf', async (req, res) => {
  //route not being called
  console.log("test");
  res.status(200).send("test");
  try {

    const game = await getGame(req.params.gameId);

    const game_obj = makeGameObject(game.variant, game.config);

    game.moves.forEach((moves) => {
      const { player, move } = getOnlyMove(moves);
      game_obj.playMove(player, move);
    });

    if (game_obj.getSGF() === '') {
      throw new Error(`SGF not supported for variant ${game.variant}`);
    }
    
    // download sgf
    res.download(game_obj.getSGF());
  } catch (e) {
    res.status(500).json({ error: e.message });
  }
});
```
DownloadSGF.vue
```


<script setup lang="ts">


const props = defineProps<{ gameId?: string}>();

</script>

<template>
  <a :href="`/game/${props.gameId}/sgf`">Download Game Files</a>
</template>
  
  <style scoped>
  button {
    margin-top: 10px;
  }
  </style>
  ```

File: packages/vue-client/src/components/GameView/NavButtons.vue [hunk 2]
```diff
@@ -33,9 +34,18 @@ const navigateRounds = (offset: number): void => {
       <button @click="() => navigateRounds(10)">>></button>
       <button @click="() => navigateRounds(props.gameRound ?? 0)">>>></button>
     </div>
+
+    
   </div>
+
+  <DownloadSGF />
```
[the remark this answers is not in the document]
SameerDalal: Fixed!

packages/vue-client/src/components/GameView/DownloadSGF.vue
SameerDalal: Fixed!

benjaminpjones:
> Whenever the download button is clicked the link is changes to /game/{gameID}/sgf, which is correct. However in api.ts, the router.get function, which executes the download, is not being called.

Ah, I forgot - we prepend `/api` to all the routes in api.ts.  If you do /api/game/{gameID}/sgf, the route will be called.

---

There is another issue once you get that fixed.  `res.download()` takes a filepath, not the file content, which is why I recommend trying that pipe stuff from this [SO thread](https://stackoverflow.com/questions/45922074/node-express-js-download-file-from-memory-filename-must-be-a-string):

```
    var fileContents = Buffer.from(sgfString, "base64");
    var fileName = `${req.params.gameId}.sgf`
  
    var readStream = new stream.PassThrough();
    readStream.end(fileContents);

    response.set('Content-disposition', 'attachment; filename=' + fileName);
    response.set('Content-Type', 'text/plain');

    readStream.pipe(response);
```

SameerDalal:
Thank you for your help! The download SGF feature should be complete and is ready to be merged into the main branch.

packages/shared/src/variants/__tests__/quantum.test.ts
```diff
@@ -182,6 +182,8 @@ test("Two passes ends the game", () => {
 
   game.playMove(0, "pass");
   game.playMove(1, "pass");
+  
+  expect(game.getSGF()).toBe("(;\nEV[GO Variants]\nPB[player Black]\nPW[player white]\nSZ[4:2]\nKM[6.5]\nVS[quantum]\n\n\n;B[ba];W[ca];B[bb];W[cb]\n\n)");
```
benjaminpjones, line 186: Ah I think I understand your comment on the other PR now

> Are there any benefits to creating a more complex test?

I had written a standalone test, which has some benefits, but this is fine too :)

packages/server/src/api.ts
```diff
@@ -26,6 +28,39 @@ import { io } from "./socket_io";
...
+    game.moves.forEach((moves) => {
+      const { player, move } = getOnlyMove(moves);
+      game_obj.playMove(player, move);
+    });
+
+    if (game_obj.getSGF() === '') {
+      throw new Error(`SGF not supported for variant ${game.variant}`);
+    } else if (game_obj.getSGF() === "non-rectangular"){
+      throw new Error('SGF for non rectangular boards is not possible');
+    } else if (!(game_obj.phase === 'gameover')){
```
benjaminpjones, line 48: For my own understanding is there something about SGF that requires the game to be finished? This is totally fine regardless, just wondering if we should consider implementing mid-game sgf download in the future.

benjaminpjones (approved):

benjaminpjones:
Nice work, thank you Sameer! I'll test a bit then merge later tonight or tomorrow

packages/shared/src/variants/__tests__/quantum.test.ts
SameerDalal: Perfect!

packages/server/src/api.ts
SameerDalal: We could definitely implement a mid-game sgf download if needed; it wouldn't require that much modification.

SameerDalal: > We could definitely implement a mid-game sgf download if needed; it wouldn't require that much modification.



> We could definitely implement a mid-game sgf download if needed; it wouldn't require that much modification.

Now that I think about that, our current implementation relies on the game to be over before it saves the entire sgf. It is still possible but would require a different approach

benjaminpjones:
When merging, I noticed lints and tests failed.  Here are the commands to run them locally:

```
yarn lint
yarn test
```

I have a [precommit hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) set up to run these before every commit, so I don't have to think about it.  Pretty easy to set up - just add a file named `.git/hooks/pre-commit` with the following:

```
yarn lint:check
yarn test
```

benjaminpjones:
Changes are live!

PR #274, opened 2024-06-12
Revert "Revert "send state via api instead of all moves""

merowin:
Reverts govariantsteam/govariants#271


benjaminpjones (approved):

benjaminpjones:
Thanks for fixing!

PR #276, opened 2024-06-13
merge sgf download to main

SameerDalal:


benjaminpjones:
Does this mean we should close #273?

SameerDalal:
> Does this mean we should close #273?

Just closed it! However, it seems that none of my commits are properly being built by github. I'll have to take a look at that before merging.

packages/shared/src/abstract_game.ts
```diff
@@ -15,6 +15,11 @@ export abstract class AbstractGame<GameConfig = object, GameState = object> {
     this.config = config ?? this.defaultConfig();
   }
 
+  /**
+   * Returns the SGF data for a game
+   */
+  abstract getSGF(): string;
```
benjaminpjones, line 21: Let's make this optional, for three reasons:

1) Smaller PR (we can remove many `getSGF { return "" }`)
2) Future variant implementers won't be obligated to implement immediately (new variant creation is already pretty heavy)
3) We can hide the "Download" button for variants without a proper implementation

Optional functions can be implemented via [interface merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html):

```
export interface AbstractGame<GameConfig = object, GameState = object> {
  /**
   * Returns the SGF data for a game
   */
  getSGF?: () => string;
}

export class AbstractGame<GameConfig = object, GameState = object> {
     ... existing stuff ...
}
```

packages/shared/src/variants/quantum.ts
```diff
@@ -49,9 +49,16 @@ class BadukHelper {
...
+  
+  public sgfContent: string = '';
+  private moves: string[] = [];
+  public moveNumber: number = 0;
+
 
   constructor(config: BadukConfig) {
     super(isLegacyBadukConfig(config) ? mapToNewConfig(config) : config);
+    this.bootstrap("player Black","player white", 9, 6.5); 
+    // TODO: make sure that player names, board size, and komi arugments are dynamic with user input
```
benjaminpjones, line 61: I'm okay with hard-coding `PB` and `PW` for now, but can we implement size and komi?  Both are available on `BadukConfig`, so it should be trivial to pass into `bootstrap()`.

Note we support rectangular boards and so does SGF, so `SZ` might look something like `SZ[${width}:${height}]`.

packages/shared/src/variants/quantum.ts
```diff
@@ -49,9 +49,16 @@ class BadukHelper {
...
   quantum_stones: Coordinate[] = [];
+  
+  public sgfContent: string = '';
+  private moves: string[] = [];
+  public moveNumber: number = 0;
+
 
   constructor(config: BadukConfig) {
     super(isLegacyBadukConfig(config) ? mapToNewConfig(config) : config);
+    this.bootstrap("player Black","player white", 9, 6.5);
```
benjaminpjones, line 60: nit: can we name this `generateSGFHeader()` or something?  `bootstrap()` could be just about anything.

packages/vue-client/src/router/index.ts
```diff
@@ -31,6 +31,12 @@ const router = createRouter({
       component: () => import("../views/GameView.vue"),
       props: true,
     },
+    {
```
benjaminpjones, line 34: I don't believe this is necessary - this file defines the Vue (UI) pages, not the Express (`/api/`) routes.

packages/vue-client/src/views/GameView.vue
```diff
@@ -197,15 +199,28 @@ const createTimeControlPreview = (
...
+
     <component
       v-if="variantGameView && game_state?.state"
       v-bind:is="variantGameView"
       v-bind:gamestate="game_state.state"
       v-bind:config="config"
       v-on:move="makeMove"
     />
+
+    <div v-if="game.result" style="font-weight: bold; font-size: 24pt; margin-bottom: 0pt" >
```
benjaminpjones, line 215: I believe these merge conflicts are what the CI is complaining about.  If you remove all the changes except the `DownloadSGF` component, it should build okay.

btw, the command is `yarn build` to build locally :)

yarn.lock
benjaminpjones, line 1: Can we revert these changes?  I think this and `package.json` changes are unrelated to the SGF download feature.

packages/shared/src/variants/quantum.ts
```diff
@@ -234,8 +250,53 @@ export class QuantumGo extends AbstractGame<NewBadukConfig, QuantumGoState> {
...
+      "\n"
+
+    ];
+  
+    this.writeToSGF(initData, true);
+   
+  }
+  
+  //creating sgf file
+  public writeToSGF(text: string[], callingBootstrap: boolean){
```
benjaminpjones, line 279: Does this get used outside this class?  If not, let's make it `private`

packages/shared/src/variants/quantum.ts
```diff
@@ -234,8 +250,53 @@ export class QuantumGo extends AbstractGame<NewBadukConfig, QuantumGoState> {
...
+        formattedMove = `${i % 2 === 0 ? ";B" : ";W"}[${text[i]}]`;
+      }
+      this.sgfContent += formattedMove;
+    }
+    if(!callingBootstrap){
+      this.sgfContent += "\n\n)";
+    }  
+  }
+
+  getSGF(): string{
```
benjaminpjones, line 293: Could we get a basic test to cover the new functionality?  QuantumGo unit tests are in [__tests__/quantum.test.ts](https://github.com/govariantsteam/govariants/blob/main/packages/shared/src/variants/__tests__/quantum.test.ts).  Something like this would be sufficient:

```
const game = new QuantumGo({
  board: { type: "grid", width: 2, height: 2 },
  komi: 5.5,
});
game.playMove(0, "aa");
game.playMove(1, "bb");
game.playMove(0, "pass");
game.playMove(1, "pass");
expect(game.getSGF()).toBe("""
 ... sgf data here ...
""")
```

Let me know if I can answer any questions about [Jest](https://jestjs.io/)

benjaminpjones (changes requested):
Awesome to see this almost ready!  Thank you so much for putting this together!

Now that the feature is fully functional, I left a more thorough code review.  Please let me know if I can help with anything, or clarify my comments!

packages/shared/src/variants/quantum.ts
merowin, line 293: Somewhere we should check whether this game is played on a grid board or not, because non-rectangular boards are unfortunately not compatible with SGF.

For example:

```
getSGF(): string {
  if (!isGridBadukConfig(this.config)) {
    // SGF for non rectangular boards is not possible
    return "";
  }
  return this.sgfContent;
}
```

merowin (commented):
Nice work, thanks!

packages/shared/src/variants/quantum.ts
SameerDalal: It seems as though BadukConfig does not have the board property. When obtaining the width using this: `config.board.width`. I get the error: " Property 'board' does not exist on type 'BadukConfig'. Property 'board' does not exist on type 'LegacyBadukConfig'."

yarn.lock
SameerDalal: I converted it back to the file when I initially forked the repo. Are those the changes that you want or the ones that are currently in the main branch of the repo?

packages/shared/src/variants/quantum.ts
SameerDalal: Are there any benefits to creating a more complex test?

packages/shared/src/variants/quantum.ts
merowin: The structure of the config has changed during the development process, that is why BadukConfig is of type LegacyBadukConfig or NewBadukConfig. We support both at the moment, because the database contains games with the old config. But we can use typeguard functions to distinguish between the cases.

How about this:

```
  constructor(config?: BadukConfig) {
    super(isLegacyBadukConfig(config) ? mapToNewConfig(config) : config);

    if (isGridBadukConfig(this.config)) {
      if (this.config.board.width >= 52 || this.config.board.height >= 52) {
        throw new Error("Baduk does not support sizes greater than 52");
      }
      this.bootstrap("player Black","player white", this.config.board.width, this.config.board.height, this.config.komi);
      [...]
    }
  }
```

packages/shared/src/variants/quantum.ts
```diff
@@ -168,6 +180,10 @@ export class QuantumGo extends AbstractGame<NewBadukConfig, QuantumGoState> {
         this.subgames[1].clear(quantum_captures[0]);
       }
     }
+    // adding the coordinates to the moves array
+    this.moves[this.moveNumber] = move;
```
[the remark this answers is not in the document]
SameerDalal: This works as well! When I was first building this I didn't realize that a new instance of the board was created for each move which resets the moves array.

packages/shared/src/variants/quantum.ts
```diff
@@ -234,8 +250,53 @@ export class QuantumGo extends AbstractGame<NewBadukConfig, QuantumGoState> {
...
+
+    const initData: string[] = [
+
+      "(;\n", 
+      "EV[GO Variants]\n", 
+      `PB[${playerB}]\n`, 
+      `PW[${playerW}]\n`,
+      `SZ[${boardSize}]\n`,
+      `KM[${komi}]\n`,     
+      `DT[${currentDate.getMonth()+1}/${currentDate.getDate()}/${currentDate.getFullYear()} ${currentDate.getHours()}:${currentDate.getMinutes()}]\n`,
```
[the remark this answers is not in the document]
SameerDalal: I see. Until we have a better way of getting the date the game was played, I will just leave the date parameter out.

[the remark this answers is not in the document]
SameerDalal: I am a bit surprised as to why this gets the date of the download though. When the download button is clicked, what is calling the makeSGFHeader function again that updates the date to the current date?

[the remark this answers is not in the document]
merowin: Most properties in the game classes are not stored in the database. Only variant, config, moves, players, and time control are persistently saved in the DB.
The endpoint that you added fetches the game data from the DB and creates a new instance of the class for this variant (e.g. QuantumGo). At this point the constructor is called, and the current date is added to the SGF.
Afterwards all moves are processed, and this is how we get the current game state, i.e. the board etc.

SameerDalal:
I reopened #273 because that is the correct branch. This PR was accidentally opened because I was editing on the wrong branch. All the change suggestions in here have been made in PR 273. Sorry about the confusion.

packages/shared/src/variants/quantum.ts
benjaminpjones: More complex?  What do you have in mind?

A simple unit test is okay with me as long as this function is covered.

yarn.lock
benjaminpjones: It should be whatever is currently on the main branch.

packages/shared/src/variants/quantum.ts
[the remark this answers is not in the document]
SameerDalal: Got it, thank you!

PR #277, opened 2024-06-16
fix board dimensions layout

### LINKED ISSUE #257: Overlapping Layout in game page for variant quantum
In the game page for variant quantum, the next-to-play-indicator overlaps the navigation buttons:

grafik

This is not the case for other variants:

grafik

merowin:
fixes #257

There were conflicting definitions of width and height of the board svg element. Note that GameView.vue has the following style definition:

```
@media (min-width: 664px) {
  .board {
    width: var(--board-side-length);
    height: var(--board-side-length);
  }
}
```

Additionally, the board implementations had width and height set to 100%, which caused the nav-buttons and variant description to overflow into the next grid element.

Previously:

grafik

After:

grafik

### DIFF
--- a/packages/vue-client/src/components/boards/BadukBoard.vue
+++ b/packages/vue-client/src/components/boards/BadukBoard.vue
@@ -40,8 +40,6 @@ function positionClicked(pos: Coordinate) {
   <svg
     class="board"
     xmlns="http://www.w3.org/2000/svg"
-    width="100%"
-    height="100%"
     v-bind:viewBox="`-1 -1 ${width + 1} ${height + 1}`"
   >
     <rect
--- a/packages/vue-client/src/components/boards/FractionalBoard.vue
+++ b/packages/vue-client/src/components/boards/FractionalBoard.vue
@@ -63,13 +63,7 @@ const stagedMove = computed(() =>
 </script>
 
 <template>
-  <svg
-    class="board"
-    xmlns="http://www.w3.org/2000/svg"
-    width="100%"
-    height="100%"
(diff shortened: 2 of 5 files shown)


benjaminpjones:
Nice catch!

Does this actually fix #257 though?  I just opened a quantum game, and see:

Screenshot 2024-06-16 at 10 10 32 AM

I think the "Result" layout issue (which *does* seem fixed here!) may be independent from the non-square board issue.

merowin:
> I think the "Result" layout issue (which _does_ seem fixed here!) may be independent from the non-square board issue.

Thanks for catching the quantum go layout! I believe the problem comes from us using a grid layout, and some board implementations taking up the entire height of the grid element. But since there are other html elements in the same grid element (the nav buttons, and the variant description), the latter elements overflow outside of the grid element boundary, and into the next grid element, where they overlap with its contents (the next-to-move-indicator, and the game result).

So I fixed it by adjusting the height in quantum board:
grafik

To be honest I'm unsure if I like the grid layout, I'm moreso used to working with a flex layout at work.

benjaminpjones:
Thanks for fixing that too!

> To be honest I'm unsure if I like the grid layout, I'm moreso used to working with a flex layout at work.

I see Flex layout more too, and I think it would be a good fit for the global layout.

Cc: @JonKo314 for thoughts since he added the original stylesheet, but I'm guessing that's just what came with the template.

JonKo314:
Yes, the original stylesheet is just the default one from Vue. In theory I think a grid layout makes more sense, but I have almost no experience using it.

PR #278, opened 2024-06-20
Elo rating

SameerDalal:
Hi Everyone!

Spoke with Prof. Chen today and the next feature we would like to develop is the ELO rating feature for quantum. Here are some of my thoughts/questions:

1). Do we want each user to have a different ELO rating for each type of variant?
2). Players can agree to make (or not to make) a game elo-rated 
3). One of the issues with the ELO ranking system is that a person can chose not to play games to maintain their high rank. To combat this, we should implement some type of activity bonus for playing at least one game per month and a consequence if they don't.
4). Players should be matched with others that have similar rank.

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -40,7 +40,10 @@
     "ts-jest": "^29.1.1"
   },
   "nodemonConfig": {
-    "watch": ["./src", "../shared/dist"],
+    "watch": [
+      "./src",
+      "../shared/dist"
+    ],
     "delay": 500
   }
 }
--- a/packages/shared/src/api_types.ts
+++ b/packages/shared/src/api_types.ts
@@ -7,6 +7,7 @@ import {
 export interface User {
   username?: string;
   id: string;
+  elo: number;
 }
 export interface GameResponse {
   id: string;


benjaminpjones:
> 1). Do we want each user to have a different ELO rating for each type of variant?

Yes, I think. Although it might be interesting to maintain an "overall" rank and see how it compares

> 2). Players can agree to make (or not to make) a game elo-rated

This is fine, but honestly I'd be okay with just rating all games too.

> 3). One of the issues with the ELO ranking system is that a person can chose not to play games to maintain their high rank. To combat this, we should implement some type of activity bonus for playing at least one game per month and a consequence if they don't.

The server not well populated yet - I don't want to "punish" people for not playing.  I would be okay with increasing the deviation parameter over time though, since accuracy does go down over time.  I believe the ELO/glicko systems allow for this.

Also are you/Zi strongly attached to Elo?  I understand [Glicko2](https://en.wikipedia.org/wiki/Glicko_rating_system) is generally considered an improvement on ELO (but still similar), and glicko2 does update RD to reflect inactivity.

> 4). Players should be matched with others that have similar rank.

Auto-match would be great! I imagine that's a feature on its own.

---

Meta-comment - we can have discussions like these in the issues (PRs are usually for code), but thank you for gathering feedback early in the process!

SameerDalal:
> The server not well populated yet - I don't want to "punish" people for not playing. I would be okay with increasing the deviation parameter over time though, since accuracy does go down over time. I believe the ELO/glicko systems allow for this. 

Got it!

> Also are you/Zi strongly attached to Elo? I understand [Glicko2](https://en.wikipedia.org/wiki/Glicko_rating_system) is generally considered an improvement on ELO (but still similar), and glicko2 does update RD to reflect inactivity.

I have no preference for what type of rating system we want to use, but I will talk to Prof. Chen about it again and see if he has some preference. 

> Meta-comment - we can probably have discussions like these in [the issues](https://github.com/govariantsteam/govariants/issues/new) (PRs are usually for code), but I appreciate you gathering feedback early in the process!

Sorry about that, I assumed I would have some code later, so I opened up a PR to keep this discussion in one section. Next time, I'll be sure to open up an issue first!

benjaminpjones:
Closing this, we can continue discussion on #279.  Feel free to open a new PR when you have some code!

(Also, I might recommend starting a new branch off current govariants/main - no need to start with a merge commit 😃)

PR #280, opened 2024-06-23
Hide download button if it's not supported

benjaminpjones:
Most variants don't support file download, so we can hide the button for them:

Screenshot 2024-06-23 at 7 39 20 PM

### DIFF
--- a/packages/vue-client/src/views/GameView.vue
+++ b/packages/vue-client/src/views/GameView.vue
@@ -2,6 +2,7 @@
 import {
   HasTimeControlConfig,
   timeControlMap,
+  game_map,
 } from "@ogfcommunity/variants-shared";
 import * as requests from "../requests";
 import SeatComponent from "@/components/GameView/SeatComponent.vue";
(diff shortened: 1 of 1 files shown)


merowin (approved):
Looks clean! I'm not very familiar with prototype, and I thought the entries of game_map are functions 😅 
I assume this condition checks whether the prototype for the variant class has the getSGF property.

SameerDalal (approved):
Looks good! Not exactly sure how prototype works, but if getSGF is undefined or some null equivalent, then the statement will evaluate to false.

benjaminpjones:
Thanks for the reviews!

> I'm not very familiar with prototype

> Not exactly sure how prototype works

I think this is the first time I've ever accessed it directly in code, but this seems to be the JS way to get properties from the class, instead of an instance.  It's a little unfortunate - I prefer Python where the function is defined directly on the class.

> I thought the entries of game_map are functions 😅

They are! We should think of them as classes, but straight from the MDN docs: ['Classes are in fact "special functions"'](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#defining_classes).

> I assume this condition checks whether the prototype for the variant class has the getSGF property.

> if getSGF is undefined or some null equivalent, then the statement will evaluate to false.

Correct! For most variants, X.prototype.getSGF is undefined.

Issue #281, opened 2024-06-24
Quantum on Graph boards is broken


merowin:
For some reason, the board doesn't show (anymore): 

![grafik](https://github.com/govariantsteam/govariants/assets/77507448/26f90bc3-d80c-4137-8c84-e3ff38d9a434)

PR #282, opened 2024-06-24
fix layout for quantum with graph board

merowin:
fixes: https://github.com/govariantsteam/govariants/issues/281

I believe I unfortunately broke this with the pr: #277


benjaminpjones (approved):
I was wondering where those went 😆 - thanks for fixing!

PR #283, opened 2024-07-03
simpler implementation of sgf download

SameerDalal:
Code is much more simple and allows the user to download the file at any point in the game.


benjaminpjones (approved):
Looks like a nice improvement, thanks!

PR #284, opened 2024-07-28
Refactor game_map to variant_map

benjaminpjones:
Sorry for the bigger PR, but I wanted to show the final destination, and these changes needed to propagate to all variants.

## Motivation

Naturally, there are some properties of variants that do not depend on game state.  We have dealt with this in two ways:

- Create an additional data structure to index some property based on variant name
    - Examples: descriptions, deprecated variants
    - Downside: information for a variant is not colocated.  This can make it difficult for a variant implementer to know completely what needs to be added.  It's also harder to do codegen.
- Add a pseudo-static method on the `AbstractGame` class:
    - Example: defaultConfig
    - Downside: Need to construct the game object in order to access these methods.  This pattern was confusing and bug-prone.

## The change

This PR adds the `Variant` class, which holds all "stateless" information about the variant.  Currently, this consists of:

```
export interface Variant<ConfigT extends object = object> {
  gameClass: new (config: ConfigT) => AbstractGame;
  description: string;
  deprecated?: boolean;
  defaultConfig: () => ConfigT;
}
```

In the future, this I could see this including properties/methods for sanitizing legacy config entries, long-form descriptions, a flag for experimental variants etc.


JonKo314 (approved):
Thanks, looks good

Issue #270, opened 2024-06-10
Quantum go uses Coordinate.toSgfRepr() (in exportState), which throws an error with the grid boards.


merowin:
The function toSgfRepr() throws an error when the coordinate is out of its bounds, which can easily happen with the graph boards, since they use only the x-part of the coordinate.

I'm thinking we can fix it by avoiding the sgf conversion in case of a graph board.

merowin:
Example: https://www.govariants.com/game/669e8cd367a4f92670d72157

PR #285, opened 2024-07-30
Quantum bug fixes

benjaminpjones:
A couple bugs:

- I introduced a bug in #284  where specialMoves threw an error during the quantum phase
- #270: toSgfRepr throwing error on non-grid boards with coordinates greater than 52


merowin (approved):
Nice work, thank you for fixing this!

PR #288, opened 2024-09-21
Sgf for Baduk and its subclasses

benjaminpjones:
This diff introduces a class dedicated to SGF, and adds it to the Baduk variant.  This also gives Baduk subclasses SGF capabilities.

I am happy to report that SGF seems to tolerate changes in player order, so variants like Keima are able to be parsed by other programs like OGS.

Screenshot_20240921_130643

### DIFF
--- a/packages/shared/src/lib/__tests__/sgf_recorder.test.ts
+++ b/packages/shared/src/lib/__tests__/sgf_recorder.test.ts
@@ -3,14 +3,14 @@ import { SgfRecorder } from "../sgf_recorder";
 test("SgfRecorder square board", () => {
   const detector = new SgfRecorder({ width: 9, height: 9 }, 7.5);
   expect(detector.sgfContent).toBe(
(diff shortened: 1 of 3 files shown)


SameerDalal:
That's awesome! Since some of the Baduk subclasses have different board shapes (non square boards) how are you able to calculate the SGF information for those games?

SameerDalal (commented):
Looks good!

benjaminpjones:
Thanks!

> Since some of the Baduk subclasses have different board shapes (non square boards) how are you able to calculate the SGF information for those games?

I took the same approach as Quantum - non-square rectangular boards are supported e.g. `SZ[9:19]`, but other graphs will just return some "non-rectangular boards not supported" file.

I might follow up w/ a PR to hide the download button for non-rectangular configs though.

SameerDalal:
Got it!

PR #293, opened 2024-10-05
indicator of last moves in variant fractional

merowin:
Adds an indicator of last rounds moves in variant fractional. Example:

grafik


JonKo314:
Looks a bit strange to me. Can we make it smaller or a ring (by using `stroke` instead of `fill`)?

merowin:
Yeah, that is indeed easier on the eyes:

grafik

benjaminpjones (approved):

JonKo314 (approved):

PR #294, opened 2024-10-12
unsub and subscribe to new seat in one api call

merowin:
This is an attempt to fix the bug mentioned in [https://forums.online-go.com/t/game-setup-parallel-fractional-go-game-4/53321/39](url)

I've experienced this bug on the live site too, but could not reproduce locally. So I'm not sure if this will indeed fix the bug.

In any case, I believe this is an improvement because it reduces the number of api calls when a player selects a seat. Previously two calls were triggered, one to unsubscribe and one to subscribe to the new seat. The above mentioned error could happen if the unsubscribe call is slower than the subscribe call.

### DIFF
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -133,6 +133,16 @@ io.on("connection", (socket) => {
       await socket.leave(topic);
     }
   });
+
+  socket.on("select_seat", async ([game_id, seat, nr_of_players]) => {
+    // unsubscribe from all seats
+    for (let i = 0; i < nr_of_players; i++) {
+      await socket.leave(`game/${game_id}/${i}`);
+    }
+
+    // subscribe to new seat
+    await socket.join(`game/${game_id}/${seat}`);
+  });
 });
 
 // If production, serve the React repo!
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -134,7 +134,7 @@ io.on("connection", (socket) => {
     }
   });
 
-  socket.on("select_seat", async ([game_id, seat, nr_of_players]) => {
+  socket.on("unique_subscribe", async ([game_id, seat, nr_of_players]) => {
     // unsubscribe from all seats
     for (let i = 0; i < nr_of_players; i++) {
       await socket.leave(`game/${game_id}/${i}`);
--- a/packages/vue-client/src/views/GameView.vue
+++ b/packages/vue-client/src/views/GameView.vue
@@ -126,7 +126,7 @@ const setPlayingAs = (seat: number) => {
   }
 
   playing_as.value = seat;
-  socket.emit("select_seat", [props.gameId, seat, players.value.length]);
+  socket.emit("unique_subscribe", [props.gameId, seat, players.value.length]);
 };
 
 function makeMove(move_str: string) {
--- a/packages/vue-client/src/views/GameView.vue
+++ b/packages/vue-client/src/views/GameView.vue
@@ -114,28 +114,19 @@ const leave = (seat: number) => {
     });
 };
 
-function unsubscribeAllSeats(): void {
-  if (players.value) {
-    socket.emit(
-      "unsubscribe",
-      players.value.map((_, index) => `game/${props.gameId}/${index}`),
-    );
-  }
-}
-
 const setPlayingAs = (seat: number) => {
-  if (!user.value) {
+  if (!user.value || !players.value) {
     return;
   }
-  unsubscribeAllSeats();
+
   if (playing_as.value === seat) {
     playing_as.value = undefined;
+    socket.emit("unsubscribe", [`game/${props.gameId}/${seat}`]);
     return;
   }
-  if (players.value && players.value[seat]?.id === user.value.id) {
-    playing_as.value = seat;
-    socket.emit("subscribe", [`game/${props.gameId}/${seat}`]);
-  }
+
+  playing_as.value = seat;
+  socket.emit("select_seat", [props.gameId, seat, players.value.length]);
 };
 
 function makeMove(move_str: string) {


benjaminpjones:
I think you're on the right track, but improving the way we unsubscribe from other seats (`game/{seat}`) implies that the bug is caused by the user *changing* seats.  I don't think that can happen in our Fractional game since each user has a unique seat.  If I'm interpreting the network traffic correctly, the problem is that we never unsubscribe from the game as an observer (`seat == null`).

Screenshot 2024-10-12 at 8 15 41 AM

I do think the atomic sub/unsub is good though.  Also, we might consider adding a check on the client that the message is associated with the right seat so that the client can be resilient to server issues (perhaps the server never receives a certain message)

packages/server/src/index.ts
```diff
@@ -133,6 +133,16 @@ io.on("connection", (socket) => {
       await socket.leave(topic);
     }
   });
+
+  socket.on("select_seat", async ([game_id, seat, nr_of_players]) => {
```
benjaminpjones, line 137: Nit: "select_seat" doesn't really communicate how this is different from "subscribe".  If it were me, I might call it something like "unique_subscribe" or something.

Also, do we any longer have a need for "subscribe"?  Should we deprecate it?

benjaminpjones: I just checked, and [QuantumGo_Engine](https://github.com/SameerDalal/QuantumGo_Engine/blob/main/web_board.py#L162-L164) is not reliant on the socket directly, so no worries about breaking them either.

benjaminpjones, line 137: Why is the client responsible for telling us `nr_of_players`?  Shouldn't the server know this?

merowin:
> I think you're on the right track, but improving the way we unsubscribe from other seats (`game/{seat}`) implies that the bug is caused by the user _changing_ seats. I don't think that can happen in our Fractional game since each user has a unique seat. If I'm interpreting the network traffic correctly, the problem is that we never unsubscribe from the game as an observer (`seat == null`).

Hm I think the function `setPlayingAs` is called when a user selects a seat while they already occupy the seat. The "sit down on a seat" api-call unfortunately does not subscribe the user to a socket room, because I did not know how to do that.

My thinking is that, if the unsubscribe-call is slower than the subscribe call, then the subscription created by the subscribe call is immediately destroyed by the unsubscribe-call, if it arrives at the server late. This could explain the bug, but I'm not 100% sure.

It could be caused by the observer state, but in theory state_map can hold both, because its indexed by a combination of round + (seat | null). As long as the states cached in state_map are in sync with the server game state, then its just a question of correctly displaying the state.

packages/server/src/index.ts
merowin: This is to avoid that the server needs to fetch the game from the data base. Although now that I think about it, without server side validation this could be abused to avoid the unsubscription.

packages/server/src/index.ts
```diff
@@ -133,6 +133,16 @@ io.on("connection", (socket) => {
       await socket.leave(topic);
     }
   });
+
+  socket.on("unique_subscribe", async ([game_id, seat, nr_of_players]) => {
```
merowin, line 137: renamed

benjaminpjones: Ah sorry for the confusion, I had deleted my comment around this, as I realized the original motivation better (`subscribe` is generic, `take_seat` is directly related to seats).  Either way is fine with me now, but feel free to push back if I ever comment something that doesn't make sense haha

benjaminpjones:
>  Hm I think the function setPlayingAs is called when a user selects a seat while they already occupy the seat.

Oh I got it!  I thought the problem was

1) subscribe(seat 0)
2) unsubscribe(seat 1)

But the problem you're addressing is:

1) subscribe(seat 0)
2) unsubscribe(seat 0)

Thanks for explaining!

---

As an aside, I did look it up, and websocket/socket.io *should* preserve order: https://socket.io/docs/v4/delivery-guarantees#message-ordering

So, I'm not confident pushing the logic to the server will change the order of the socket.joins.  I'm okay either way, but if it's not going to change the behavior, we should optimize for readability or whatever else.

packages/server/src/index.ts
benjaminpjones: Ah yeah, I'm not so concerned about abuse (clients still have access to generic `subscribe`), just that there is some additional complexity introduced to the client

merowin: Should I make it so the server fetches the game object from db and gets the number of players that way?

packages/server/src/index.ts
merowin: No problem, what do you think about the name "subscribe_to_unique_seat"? :-)

merowin:
The bug happens inconsistently, or at least I have not figured out a pattern yet. Since I haven't been able to reproduce the bug locally, my theory is that it's caused by some asynchronous processes. Maybe a network call takes more time than usual, or something like that.

packages/server/src/index.ts
benjaminpjones: Ohh, I think a db hit is a fine reason to do it the way you have, I'll mark this resolved!

packages/server/src/index.ts
benjaminpjones: Works for me!

benjaminpjones:
Okay, having gotten some eyes on `GameView`, I have some qualms with the way we're doing reactivity (on `main`).

Namely:

- never unsubscribing from the "unseated" topic (is this the source of our bug?)
- unsubscribing from all seats, when we really should unsubscribe from one (not a bug, but just feels like a smell)
- Too many effects in each `watchEffect`, meaning effects that should only run once are run multiple times.

See https://github.com/govariantsteam/govariants/pull/297.  I'm wondering are these the types of things worth fixing?  And if so, would they be compatible with this PR?

Let me know your thoughts!

merowin:
> Let me know your thoughts!

To be honest I don't have much experience with vues reactivity (or any reactive UI framework). I think we should merge your changes and check if they fix the bug! So I've converted this to draft for the time being.

PR #295, opened 2024-10-12
block game creation unless user is logged in

### LINKED ISSUE #147: It's possible to create a game without being logged in
We shouldn't let people do this - what's the point of creating a game you can't play?

merowin:
closes #147


benjaminpjones (approved):
Much needed fix, thanks!

Issue #112, opened 2023-06-28
Variant pages


benjaminpjones:
Currently, the only option is to link to an ongoing game.

For example, `govariants.com/variants/baduk`

Each page could have a blurb on the rules, screenshots, and a create game form.  Maybe instead of screenshots, an analysis board (see #13) and/or example game.

merowin:
I would like to work on this in future, and have some ideas on how this could be structured. What do you think of the following?

New page available from nav-bar called "Tutorial" or maybe "Rules"

Users can select a variant that they want to learn about. Then the page contains rules explanations and demo boards where users can try it out by playing moves.

The rules explanations could be divided into modules that are potentially reusable by multiple variants (and have their own components). For example a module about parallel moves could be included in both pages for Parallel and for Fractional.

Nice-to-have: cross-reference links between module sections

We'll need to keep these updated though, whenever we change the rules of a variant.

JonKo314:
Sounds like a lot of work, but also very nice!

benjaminpjones:
Awesome! I think that plan sounds good to me :)

When I wrote this issue, all I had in mind was a page with:

- Description and/or links to resources like Sensei's library
- game creation form (shouldn't be hard to pull this from the home page)
- example game (more work, but could be useful to have these for testing anyway)

I didn't think much about how it should be linked from the main page, but menu item, as you suggested, sounds nice. Or potentially a link near the game creation form on the home page after you've selected the variant.

merowin:
Ah, so each variant should get their own page then. I was thinking of one page where users can select a variant, but that wouldn't be easily linked.

benjaminpjones:
That's what I had in mind at least, but if it's preferable to have all this on one page, maybe we can support linking with query params?

`https://govariants.com/rules?variant=quantum`

And that would automatically set the selector to "quantum".

merowin:
I'm not sure if its preferable, but I think it should be one page that's linked from the navbar. We could use a similar structure as this site: https://gamerules.com/browse-all-rules/

mgrider:
Is someone is still working on this? 

Here's a checklist for which variants have rules pages that I could find just by trying to plug the name into the URL. I also included the short description for each one.

- [ ] Baduk
    Traditional game of Baduk a.k.a. Go, Weiqi
    Surround stones to capture them
    Secure more territory + captures to win
- [ ] Capture
    Baduk but the first player who captures a stone wins
- [ ] Chess
    Baduk with different types of stones
[...]

PR #296, opened 2024-10-12
add rules view

merowin:
I figured it would be good to add a rule description page. 

issue: #112 

Adds a new view for variant rule descriptions, alongside the rules of fractional.


merowin:
I don't understand why the pipeline has an error, to be honest.

benjaminpjones:
> I don't understand why the pipeline has an error, to be honest.

I think there are some HTML issues:

- p is a "self-closing" tag
- ul/li not allowed in p

https://stackoverflow.com/a/60344917

merowin:
> > I don't understand why the pipeline has an error, to be honest.
> 
> I think there are some HTML issues:
> 
>     * p is a "self-closing" tag
> 
>     * ul/li not allowed in p
> 
> 
> https://stackoverflow.com/a/60344917

Thank you, I did not know that.

packages/vue-client/src/components/RuleDescriptions/FractionalRules.vue
```diff
@@ -0,0 +1,60 @@
...
+  <h2>Base Setup</h2>
+
+  <p>
+    Fractional is a multiplayer variant for any even number of participants.
+    Players are grouped into two competing teams:
+  </p>
+  <ul>
+    <li><strong>Team Black</strong></li>
+    <li><strong>Team White</strong></li>
+  </ul>
```
JonKo314, line 11: That description is too specific to use it for *rules*, it might be better to use it as an example. The variant also allows for other colors to be primary colors. For example there could be three teams (red, green, blue) of three players each (cyan, magenta, yellow).

Here's a suggestion for a different phrasing, but feel free to change it:

> Fractional Go is a multiplayer variant with multicolored stones. Usually players are grouped into two competing teams:

packages/vue-client/src/components/RuleDescriptions/FractionalRules.vue
```diff
@@ -0,0 +1,60 @@
...
+    on the board. If two or more players played the same move, a collision
+    occurs (see <i>Collision Handling</i>).
+  </p>
+
+  <h2>Fractional Stones</h2>
+  <p>
+    Stones have multiple colours, and the chains of each colour are determined
+    individually. As a result chains may overlap with each other. Players place
+    stones with two colours, one of them being the players Team Colour, and the
+    second colour is shared with a player from the other team.
```
JonKo314, line 29: Not necessarily

JonKo314 (commented):
The implementation of this variant is a more generalized form of what was played on the OGS forums so far. It would be nice if the rules reflected that, but it also adds more complexity. Let me know, if you want me to come up with a description on my own.

packages/vue-client/src/components/RuleDescriptions/FractionalRules.vue
merowin: Should I leave out the teams aspect? Perhaps it's sufficient to describe the objective with something like

_Players place stones of two colours, one being the **Primary Colour** and one being the **Secondary Colour**. At the end of a game, players receive points equal to the area controlled by their primary colour._

benjaminpjones:
I like the idea!  One idea (that doesn't necessarily need to be implemented in this PR):

Would be nice if the *content* were not tied to Vue since it's static, and not using Vue's reactivity.  Ideally it would be some raw HTML (or better yet, Markdown), and then we could store it in `shared/` and add it to the `variant_map`, instead of creating a separate map in `RulesView`.

packages/vue-client/src/components/RuleDescriptions/FractionalRules.vue
merowin: fixed

packages/vue-client/src/components/RuleDescriptions/FractionalRules.vue
merowin: I made the rules description more generic, to reflect the flexibility of the variant.

I like the idea of an additional explanation section, but for now I'm content with adding just the rules.

merowin:
> Would be nice if the _content_ were not tied to Vue since it's static, and not using Vue's reactivity. Ideally it would be some raw HTML (or better yet, Markdown), and then we could store it in `shared/` and add it to the `variant_map`, instead of creating a separate map in `RulesView`.

I moved the html content into shared as a string and "inject" it with `vue-html`. This way we can use variant_map, unfortunately the html formatting / highlighting is lost.

JonKo314:
> unfortunately the html highlighting / auto-completion is lost.

Because it's a TypeScript file. It could also be an HTML file, but that would make importing it more complicated.

JonKo314 (approved):

benjaminpjones:
Yeah, the lack of highlighting is a bummer. As @JonKo314 said, we could push it into an html file, and get html IDE support. On the other hand, this could be a reason to use Markdown - it's easy to read and write even without IDE support.

merowin:
I don't know how to render the markdown, but in theory it sounds good.

Issue #254, opened 2024-05-05
Sequential Fractional Baduk


merowin:
I would like to add a variant with multicolored stones (fractional) and sequential moves. As I envision it, this could be played by 2 players (or any even number for a rengo-like team game).

Both players (or teams) have a fixed primary color (Black or White). Moves alternate between Black and White, but with the secondary colors also cycling. For example with 2 secondary colors:
(Black|Red), (White|Red), (Black|Blue), (White|Blue), repeating

The number of secondary colors could also be 3, 4, etc. ...

Both players objective is to maximise the area of their primary color.

Here is a forum game thread as reference: https://forums.online-go.com/t/fractional-team-go-game-1/37048/

PR #298, opened 2024-10-13
Sequential fractional

merowin:
Add 2 player variant with fractional stones and sequential moves.

issue: #254


packages/vue-client/src/components/GameCreation/SFractionalConfigForm.vue
```diff
@@ -0,0 +1,64 @@
...
+function emitConfigChange() {
+  emit("configChanged", config);
+}
+
+function setBoardConfig(boardConfig: BoardConfig): void {
+  config.board = boardConfig;
+  emitConfigChange();
+}
+
+/* TODO: Figure out how to make this work with vues reactivity
```
benjaminpjones, line 28: I think you just need to wrap `config` in `reactive()`:

```
const config: SFractionalConfig = reactive({
  ...props.initialConfig,
  board: { ...props.initialConfig.board },
  secondary_colors: [...props.initialConfig.secondary_colors],
});
```

Vue is capable of detecting updates on reactive arrays!

benjaminpjones (commented):
It's a cool variant!  Generlly, I think it would be better to scope the changes only to what is related to sfractional, but otherwise I think this will be a nice addition!

packages/vue-client/src/components/GameCreation/ParallelGoConfigForm.vue
```diff
@@ -30,7 +30,7 @@ function emitConfigChange() {
   </form>
 </template>
 
-<style>
+<style scoped>
```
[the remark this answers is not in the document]
merowin: I agree, I got a bit carried away, sorry. I was trying to make the input fields align at the right, and discovered that the unscoped styles were applied.
I'll undo these changes, and make a separate pr for the style changes.

benjaminpjones (approved):
Thanks for the clean up! I think 14 files is more reasonable (esp. considering vue changes are mostly optional for variant implementers)

merowin:
Hm there is some weird behaviour where clicking the + button in the creation form triggers a page reload. I'll need to find out the reason.

packages/vue-client/src/assets/main.css
```diff
@@ -113,4 +113,4 @@ a,
 .info-label {
   font-weight: bold;
   margin-right: 0.5em;
-}
\ No newline at end of file
+}
```
JonKo314, line 116: Why delete the empty line at the end of that file? Was that some auto formatting?

packages/vue-client/src/components/boards/SFractional/SFractionalGridBoard.vue
JonKo314, line 1: `SFractionalGraphBoard.vue` and SFractionalGridBoard.vue` are pretty similar, I wonder if they should be put into one file.

merowin: I think the code is easier to read and cleaner with separate files in this case.

packages/vue-client/src/assets/main.css
merowin: I don't know to be honest. Should I revert it?

merowin:
> Hm there is some weird behaviour where clicking the + button in the creation form triggers a page reload. I'll need to find out the reason.

Fixed, seems like buttons in forms trigger a page reload. Maybe like a submit form button would.

packages/vue-client/src/components/GameCreation/SFractionalConfigForm.vue
```diff
@@ -45,30 +45,30 @@ function removeSecondaryColor(): void {
...
+    <label>Secondary Colors</label>
+
     <input
       type="color"
       v-for="(_, index) in config.secondary_colors"
       :key="index"
       v-model="config.secondary_colors[index]"
     />
   </form>
+  <!-- Note: I would like to place these buttons above the color inputs rather than below. -->
```
benjaminpjones, line 58: Possible fix: `type="button"`

https://stackoverflow.com/questions/7803814/how-can-i-prevent-refresh-of-page-when-button-inside-form-is-clicked

packages/vue-client/src/assets/main.css
JonKo314: Maybe I misunderstood. I think GitHub is saying the `}` without `\n` was removed and a `}\n` was added, but the newline isn't visible in GitHub. Anyway, if the auto format thinks it should be this way, there's no need to revert it.

packages/vue-client/src/components/boards/SFractional/SFractionalGridBoard.vue
JonKo314: Yes, it's easier to read. But as it is also a lot of duplicated code, changes in the future will probably need to be done twice. It's not a big deal though.

JonKo314 (approved):

merowin:
Thank you both for the review!

Issue #211, opened 2024-03-01
Display player colors or symbols


evizam:
Currently, it's hard to tell for an observer of a game which player is playing which color. For some variants (e.g. fractional go) it's even impossible.

My proposal is to add an indicator to each player seat. For example, a stone in the player's color next to the player number. To keep this "abstract", every game variant could define a "symbol" for each player (and each round, in case that can change) and pass it to the player seat.

merowin:
Since we've got a prototype that works for all current variants relatively well, I'll close this for now.

Issue #134, opened 2023-07-09
Secondary Graphical Component for Games


benjaminpjones:
For all games, we provide one graphical component: the board (see [board_map.ts](https://github.com/govariantsteam/govariants/blob/d5c83426af937c06af38d37b29495a98b3a2cf89/packages/vue-client/src/board_map.ts)).  For most variants, there are special bits of information we might want to show, but not on the main board.  For example:

- Thue-Morse, we want to show the upcoming move pattern (...WBB**W**BWW...)
- Frozen go might benefit by showing that we are in a frozen state.
   - Actually we could do this with the main board - maybe change the background to blue?
- We could replace the [specialMoves](https://github.com/govariantsteam/govariants/blob/d5c83426af937c06af38d37b29495a98b3a2cf89/packages/shared/src/abstract_game.ts#L59) method, and just move that logic to the client code.

A secondary game component would be a nice way to add game controls and information that doesn't fit nicely on the board itself.  Importantly, this component wouldn't show up on the home page, where we require a more minimal, if incomplete, representation of the games.

benjaminpjones:
closed by #305

PR #299, opened 2024-10-14
add player symbol component

merowin:
issue: #211

### DIFF
--- a/packages/vue-client/src/components/GameView/PlayerSymbol.vue
+++ b/packages/vue-client/src/components/GameView/PlayerSymbol.vue
@@ -1,0 +1,65 @@
+<script setup lang="ts">
+import { computed } from "vue";
+import TaegeukStone from "../TaegeukStone.vue";
+
+const props = defineProps<{
+  variant: string;
+  config: unknown;
+  gamestate: unknown;
+  player_nr: number;
+}>();
+
+const colors = computed(() => {
+  if (
+    [
+      "baduk",
+      "badukWithAbstractBoard",
+      "phantom",
+      "capture",
+      "tetris",
+      "pyramid",
+      "thue-morse",
+      "freeze",
+      "keima",
+      "drift",
+      "quantum",
+    ].includes(props.variant)
+  ) {
+    return [props.player_nr % 2 === 0 ? "black" : "white"];
+  }
+
+  if (props.variant === "chess") {
+    return [props.player_nr % 2 === 0 ? "white" : "black"];
+  }
+
+  if (props.variant === "fractional") {
+    // TODO: parse config to determine the players colors.
+    // unfortunately I can't figure out how to do it, because
+    // props.config is a reactive ref object. I don't understand why,
+    // because we also pass the config to the variantGameView.
+  }
+
+  return undefined;
+});
+</script>
+
+<template>
+  <svg v-bind:viewBox="`0 0 2 2`">
+    <TaegeukStone
+      v-if="colors"
+      :colors="colors"
+      :r="1"
+      :cx="1"
+      :cy="1"
+    ></TaegeukStone>
+  </svg>
+</template>
+
+<style scoped>
+svg {
+  margin: 10px;
+  width: 40px;
+  height: 40px;
+  filter: drop-shadow(3px 5px 6px rgb(0 0 0 / 0.4));
+}
+</style>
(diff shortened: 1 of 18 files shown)


packages/vue-client/src/components/GameView/PlayerSymbol.vue
```diff
@@ -0,0 +1,65 @@
...
+import TaegeukStone from "../TaegeukStone.vue";
+
+const props = defineProps<{
+  variant: string;
+  config: unknown;
+  gamestate: unknown;
+  player_nr: number;
+}>();
+
+const colors = computed(() => {
```
benjaminpjones, line 12: Could we push the color logic into the variants themselves? Benefits:

- baduk subclasses get their impl for free
- no need for extended if statements
- don't need to worry about Vue's reactivity

Something like this in AbstractGame:

```
playerColor(player: number): string[]
```

packages/vue-client/src/components/GameView/PlayerSymbol.vue
```diff
@@ -0,0 +1,65 @@
...
+
+  if (props.variant === "chess") {
+    return [props.player_nr % 2 === 0 ? "white" : "black"];
+  }
+
+  if (props.variant === "fractional") {
+    // TODO: parse config to determine the players colors.
+    // unfortunately I can't figure out how to do it, because
+    // props.config is a reactive ref object. I don't understand why,
+    // because we also pass the config to the variantGameView.
```
benjaminpjones, line 39: I am still wrapping my head around what is reactive and what's not in Vue, but I thought reactive objects can be treated mostly like non-reactive objects. The difference is that mutating them will trigger a re-render.

If you need to get a non-reactive version of the object, it looks like you can use `structuredClone` (though it seems a bit inefficient to create a proxy, then clone it back to non-proxy haha)

https://stackoverflow.com/questions/53662412/how-can-i-make-a-value-in-vue-js-non-reactive

packages/vue-client/src/components/GameView/SeatComponent.vue
```diff
@@ -15,6 +16,9 @@ defineProps<{
   time_control: IPerPlayerTimeControlBase | null;
   time_config?: ITimeControlConfig;
   is_players_turn: boolean;
+  variant: string;
+  config: unknown;
```
benjaminpjones, line 20: Since we're passing in `config` now, should we just get `time_config` from that?

packages/vue-client/src/components/GameView/PlayerSymbol.vue
```diff
@@ -0,0 +1,65 @@
+<script setup lang="ts">
+import { computed } from "vue";
+import TaegeukStone from "../TaegeukStone.vue";
+
+const props = defineProps<{
+  variant: string;
+  config: unknown;
+  gamestate: unknown;
```
benjaminpjones, line 8: what would gamestate be necessary for?

packages/vue-client/src/components/GameView/PlayerSymbol.vue
merowin: My concern is variants where the move colors change from round to round. So far we don't have such a variant, but an example is the one I've been working on, sequential fractional.

I chose this general structure because I wanted to keep all the flexibility, i.e. enable variants to define their own UI component that gets displayed.

packages/vue-client/src/components/GameView/SeatComponent.vue
merowin: Yeah that makes a lot of sense, although now I'm unsure if passing in config is necessary.

packages/vue-client/src/components/GameView/PlayerSymbol.vue
benjaminpjones: > My concern is variants where the move colors change from round to round

Oh makes sense! Though I wasn't sure if SFractional might just use the primary color as the symbol.

> i.e. enable variants to define their own UI component that gets displayed.

Makes sense, I think this is a good feature to enable (#134).  Though I hadn't envisioned this inside the SeatComponents - for the Thue-Morse example, wouldn't we just want a standalone UI component?  Something like this:

image

merowin: Yeah I agree. So would it be ok if I add this to the Variant interface:

`  getPlayerColors: (config: ConfigT, playerNr: number) => string[];`

and then display a Stone with these colors as the player symbol?

benjaminpjones: That works for me! And definitely seems like we can add gamestate later if needed for sfractional et al

merowin:
grafik

benjaminpjones (approved):
Looks great, thanks!

PR #300, opened 2024-10-14
Use markdown instead of HTML for rules

benjaminpjones:
Demo of markdown, using markdown-it (seems to be well maintained)


merowin:
This is nice. I guess my only comment is, does this structure support everything that we want to do (in future)?

For example I could imagine somebody to want to add an image etc. to the rules description / explanation. Or perhaps a link to the demo board.

benjaminpjones:
> For example I could imagine somebody to want to add an image etc. to the rules description / explanation. Or perhaps a link to the demo board.

While not as flexible as HTML, I think we at least get links and images out of the box:

```
[Link](http://a.com)
![Image](http://url/a.png)
```

`markdown-it` supports lots of extensions too, so I think we can enable things like inline HTML. Haven't played with it too much though.

merowin (approved):

PR #303, opened 2024-10-17
change order of last move indicator and click placeholder

merowin:
This fixes a problem that can occur with the last move indicator in fractional when a played stone gets immediately captured. Then previously the last move indicator was positioned in front of the click placeholder, making it impossible to play a move at that position.

To fix this, I changed the order of elements. Had to adjust the v-else to a v-if with negated condition.


benjaminpjones (approved):
Nice catch!

PR #304, opened 2024-10-19
enable graph boards for freeze

merowin:
Refactor Freeze board component to enable the graph boards.

This is an addendum to the rework where we enabled graph boards for baduk.


benjaminpjones:
Awesome, thanks! I think at some point we should try to abstract some of the duplicated logic way, i.e. have some `MulticolorBoard` that can create a grid or graph without too much effort on the Variant's side. But that would take some effort, so this PR LGTM!

benjaminpjones (approved):

PR #305, opened 2024-10-20
add playing table map

merowin:
issue #134

I added `playing_table_map` that is a copy of `board_map`, and use this new map in GameView. Notably the GamesList still uses `board_map`.

Should be no functional change yet. This PR is to discuss whether we want this structural change.

### DIFF
--- a/packages/vue-client/src/playing_table_map.ts
+++ b/packages/vue-client/src/playing_table_map.ts
@@ -1,31 +1,5 @@
 import { Component } from "vue";
-import BadukBoardSelector from "./components/boards/BadukBoardSelector.vue";
-import BadukBoardAbstract from "./components/boards/BadukBoardAbstract.vue";
-import ParallelGoBoard from "./components/boards/ParallelGoBoard.vue";
-import ChessBoard from "./components/boards/ChessBoard.vue";
-import FreezeGoBoard from "./components/boards/FreezeGoBoard.vue";
-import FractionalBoard from "./components/boards/FractionalBoard.vue";
-import KeimaBoard from "./components/boards/KeimaBoard.vue";
-import QuantumBoard from "./components/boards/QuantumBoard.vue";
-import SFractionalBoardSelector from "./components/boards/SFractional/SFractionalBoardSelector.vue";
 
 export const playing_table_map: {
   [variant: string]: Component<{ config: unknown; gamestate: unknown }>;
-} = {
-  baduk: BadukBoardSelector,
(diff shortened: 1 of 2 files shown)


merowin:
>     * Variant implementers should be able to do the bare minimum. If they only implement the primary board component, they shouldn't need to add boilerplate for the secondary component.

Admittedly, this change means that variant implementers would, at bare minimum, have to add two mappings to the board component, one in board_map, and one in playing_table_map. Maybe I should instead set it up such that, if there is no entry in the playing_table_map, then we default to the board component.

>     * What is meant to go into the "table map" eventually? Will it still be a `{ [ variant: string ]: Component }`? Are we expecting implementers to create three components altogether? 1 board, 1 secondary, and 1 to compose them?

Yes it will still be `{ [ variant: string ]: Component }`. And I imagine that for many variants, it is sufficient to have this be the board component. For variants where we want to display an extended view in GameView, this can be a different component that includes the board component. So I'm unsure why there may be a need to add a third component that composes them.

benjaminpjones:
> Maybe I should instead set it up such that, if there is no entry in the playing_table_map, then we default to the board component.

I think that would be a good approach! And preferable given that most variants don't need a "secondary component"

> For variants where we want to display an extended view in GameView, this can be a different component that includes the board component.  So I'm unsure why there may be a need to add a third component that composes them.

In my comment about 3 components, I was thinking it had to be something like this:

```
<TableComponent>
    <MyBoard />
    <SecondaryComponent />
</TableComponent>
```

But you're right, there's no reason the secondary component can't simply be defined in the table component.  Though I think we will likely want to break out the SecondaryComponent for code re-use anyway:

```
<ParallelTableComponent>
    <ParallelGoBoard />
    <PendingMovesComponent />
<ParallelTableComponent />

<FractionalTableComponent>
    <FractionalBoard />
    <PendingMovesComponent />
<ParallelTableComponent />
```

benjaminpjones (approved):
Makes sense, thanks!

packages/vue-client/src/playing_table_map.ts
```diff
@@ -0,0 +1,5 @@
+import { Component } from "vue";
+
+export const playing_table_map: {
```
benjaminpjones, line 3: Possibly add comment here describing the purpose of the components here, and the relationship to board_map

merowin:
should I undo the renaming of the computed variable to `variantGameView`?

PR #307, opened 2024-10-21
Use badukVariant values by default.

benjaminpjones:
Problem: lots of this:

```
  defaultConfig: Baduk.defaultConfig,
  getPlayerColors: Baduk.getPlayerColors,
  // In the future
  sanitizeConfig: Baduk.sanitizeConfig,
  anotherProperty: Baduk.anotherProperty,
```

Not sure if this is the way, but I'm looking to prevent a bunch of copied code for Baduk-based variants.  Would love your thoughts.


merowin (approved):
Yeah, I think this is a nice improvement!

PR #309, opened 2024-10-22
Sanitize config as soon as it comes from db.

benjaminpjones:
Currently, we sanitize configs *inside* the Game objects, but there are other places that access configs.  (e.g. GameView).  Let's sanitize outwardFacingGame so that we can always assume we have a modern config.

### DIFF
--- a/packages/shared/src/variant_map.ts
+++ b/packages/shared/src/variant_map.ts
@@ -70,6 +70,17 @@ export function getDefaultConfig(variant: string): object {
   return variant_map[variant]?.defaultConfig();
 }
 
+export function sanitizeConfig(
+  variant: string,
+  config: object,
+): object | undefined {
+  const sanitize = variant_map[variant]?.sanitizeConfig;
+  if (sanitize == null) {
+    return config;
+  }
+  return sanitize(config);
+}
+
 export function supportsSGF(variant: string) {
   return Boolean(variant_map[variant]?.gameClass.prototype.getSGF);
 }
--- a/packages/shared/src/variants/baduk.ts
+++ b/packages/shared/src/variants/baduk.ts
@@ -54,7 +54,7 @@ export class Baduk extends AbstractGame<NewBadukConfig, BadukState> {
   protected sgf?: SgfRecorder;
 
   constructor(config: BadukConfig) {
-    super(isLegacyBadukConfig(config) ? mapToNewConfig(config) : config);
+    super(Baduk.sanitizeConfig(config));
 
     if (isGridBadukConfig(this.config)) {
       if (this.config.board.width >= 52 || this.config.board.height >= 52) {
@@ -252,6 +252,15 @@ export class Baduk extends AbstractGame<NewBadukConfig, BadukState> {
         return [];
     }
   }
+
+  static sanitizeConfig(config: unknown): NewBadukConfig {
+    const badukConfig = config as BadukConfig;
+    return (
+      isLegacyBadukConfig(badukConfig)
+        ? mapToNewConfig(badukConfig)
+        : badukConfig
+    ) as NewBadukConfig;
+  }
 }
 
 /** Returns true if the group containing (x, y) has at least one liberty. */
@@ -282,10 +291,13 @@ export class GridBaduk extends Baduk {
   }
 }
 
-export const badukVariant: Variant<BadukConfig> = {
+export const badukVariant: Variant<NewBadukConfig> = {
   gameClass: Baduk,
   description:
     "Traditional game of Baduk a.k.a. Go, Weiqi\n Surround stones to capture them\n Secure more territory + captures to win",
   defaultConfig: Baduk.defaultConfig,
   getPlayerColors: Baduk.getPlayerColors,
+  sanitizeConfig: Baduk.sanitizeConfig,
 };
+
+export const gridBadukVariant = badukVariant as Variant<NewGridBadukConfig>;
--- a/packages/shared/src/variants/baduk.ts
+++ b/packages/shared/src/variants/baduk.ts
@@ -255,11 +255,9 @@ export class Baduk extends AbstractGame<NewBadukConfig, BadukState> {
 
   static sanitizeConfig(config: unknown): NewBadukConfig {
     const badukConfig = config as BadukConfig;
-    return (
-      isLegacyBadukConfig(badukConfig)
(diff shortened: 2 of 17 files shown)


packages/shared/src/variants/baduk.ts
```diff
@@ -282,10 +291,13 @@ export class GridBaduk extends Baduk {
...
+export const badukVariant: Variant<NewBadukConfig> = {
   gameClass: Baduk,
   description:
     "Traditional game of Baduk a.k.a. Go, Weiqi\n Surround stones to capture them\n Secure more territory + captures to win",
   defaultConfig: Baduk.defaultConfig,
   getPlayerColors: Baduk.getPlayerColors,
+  sanitizeConfig: Baduk.sanitizeConfig,
 };
+
+export const gridBadukVariant = badukVariant as Variant<NewGridBadukConfig>;
```
merowin, line 303: I thought about whether we can improve this, and have come up with the following:
```
export const badukVariant: Variant<NewBadukConfig> & {defaultConfig: () => NewGridBadukConfig} = {
  gameClass: Baduk,
  description:
    "Traditional game of Baduk a.k.a. Go, Weiqi\n Surround stones to capture them\n Secure more territory + captures to win",
  defaultConfig: Baduk.defaultConfig,
  getPlayerColors: Baduk.getPlayerColors,
  sanitizeConfig: Baduk.sanitizeConfig,
};

export const {sanitizeConfig: _, ...gridBadukVariant} = badukVariant;
gridBadukVariant satisfies Variant<NewGridBadukConfig>;
```

However I'm not sure if this is really an improvement. On the plus side, I believe this would be more typesafe.

packages/shared/src/variants/baduk_utils.ts
```diff
@@ -49,13 +49,17 @@ export function isLegacyBadukConfig(
   );
 }
 
-export function mapToNewConfig(config: LegacyBadukConfig): NewBadukConfig {
+export function mapToNewConfig<ConfigT extends LegacyBadukConfig>(
```
merowin, line 52: Hm is there a reason against using a simpler signature?

```
export function mapToNewConfig(
  config: LegacyBadukConfig,
): LegacyBadukConfig & NewGridBadukConfig {
  return {
    ...config,
    board: {
      type: BoardPattern.Grid,
      width: config.width,
      height: config.height,
    },
  };
```

Technically speaking I think the type error thrown by transpiler is correct. If I write `mapToNewConfig<LegacyBadukConfig & {someNewProp: string}>(config)` then the function signature does not check out.

merowin (commented):
Overall a good addition, thanks!

I'm still thinking of using some sort of data base migrations, or similar, to map old configs to the new structure.

packages/shared/src/variant_map.ts
```diff
@@ -70,6 +70,17 @@ export function getDefaultConfig(variant: string): object {
   return variant_map[variant]?.defaultConfig();
 }
 
+export function sanitizeConfig(
+  variant: string,
+  config: object,
+): object | undefined {
+  const sanitize = variant_map[variant]?.sanitizeConfig;
+  if (sanitize == null) {
```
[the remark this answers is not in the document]
benjaminpjones: Yeah, I usually just do `== null` if `undefined` and `null` should be treated equally, but we could also trust TypeScript and do `=== undefined`.

BTW, just realized  [`eqeqeq`](https://eslint.org/docs/latest/rules/eqeqeq) isn't enabled by eslint's recommended config! We should turn that on in some form ("always" or "smart" would be okay with me)

packages/shared/src/variants/baduk.ts
benjaminpjones: Oh that's interesting! I am not so familiar with `satisfies`, but it definitely seems better than `as`. I do assume we need to sanitize grid variants.  Let me play around a bit and see if we can do both.

(though maybe the right move is to just purge old games...)

packages/shared/src/variants/baduk_utils.ts
benjaminpjones: I think I was trying to capture the following:

```
const config = {
      komi: 6.5,
      width: 19,
      height: 19,
      xShift: 0,
      yShift: 1,
}
const sanitizedConfig = mapToNewConfig(config);
sanitizedConfig.xShift;
                ^^^^^^
Property 'xShift' does not exist on type 'NewGridBadukConfig'.
```

But it occurs to me that `DriftGoConfig` inherits from `NewGridBadukConfig`, so we don't really make use of this complex typing.   I'll go ahead and remove it

packages/shared/src/variants/baduk.ts
```diff
@@ -252,6 +252,15 @@ export class Baduk extends AbstractGame<NewBadukConfig, BadukState> {
...
     }
   }
+
+  static sanitizeConfig(config: unknown): NewBadukConfig {
+    const badukConfig = config as BadukConfig;
+    return (
+      isLegacyBadukConfig(badukConfig)
+        ? mapToNewConfig(badukConfig)
+        : badukConfig
+    ) as NewBadukConfig;
```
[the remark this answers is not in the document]
benjaminpjones: Good catch! This wqas in here because earlier I hadn't cast config to `BadukConfig`, which broke the type deduction.

benjaminpjones:
Thanks for the thoughtful review and comments about types!

> I'm still thinking of using some sort of data base migrations, or similar, to map old configs to the new structure.

Sounds good!  I think this function could be useful for that as well - we could write a node script that pulls games from db and passes it through this sanitizeConfig function.

packages/shared/src/variants/baduk_utils.ts
merowin: Ah in that case maybe the "fix" is to change the type of the input config to ConfigT:
```
export function mapToNewConfig<ConfigT extends LegacyBadukConfig>(
  config: ConfigT,
): ConfigT & NewGridBadukConfig {
  return {
    ...config,
    board: {
      type: BoardPattern.Grid,
      width: config.width,
      height: config.height,
    },
  };
```

merowin (approved):

packages/shared/src/variants/baduk_utils.ts
benjaminpjones: Of course!  I totally missed that haha

Fixed in latest commit

Issue #137, opened 2023-07-09
Make boards more consistent


benjaminpjones:
Currently, there are 3-4 ways to make a standard looking Baduk board:

- [BadukBoard](https://github.com/govariantsteam/govariants/blob/235a56fb7520e7b67c355b4d98a497350c72c51b/packages/vue-client/src/components/boards/BadukBoard.vue) (obviously)
- [BadukBoardAbstract](https://github.com/govariantsteam/govariants/blob/235a56fb7520e7b67c355b4d98a497350c72c51b/packages/vue-client/src/components/boards/BadukBoardAbstract.vue)
- [MulticolorGridBoard](https://github.com/govariantsteam/govariants/blob/235a56fb7520e7b67c355b4d98a497350c72c51b/packages/vue-client/src/components/boards/MulticolorGridBoard.vue)
- [Coming soon] [FractionalBoard](https://github.com/govariantsteam/govariants/blob/3dcbd595acca57f9811967abdcb47d0aa53483c9/packages/vue-client/src/components/boards/FractionalBoard.vue)

<img width="615" alt="Screenshot 2023-07-09 at 1 47 55 PM" src="https://github.com/govariantsteam/govariants/assets/25233703/991a0e0d-091d-4638-a11a-e7feb400fb5c">

Some features are different by design:
 - graph vs grid
 - fractional stones

But some features are mismatched for no good reason (I'm not sure this is a complete list, but this is everything I noticed):
- last move marker
- score markings
- stone outlines
- stone width
- hover indicator
- margin width
- line weight

The task is to make the boards more consistent as the differences are quite obvious when they occur together on the game page.  There may also be room for deduping the classes.  `BadukBoard` may be implemented as a specialization of `BadukBoardAbstract` or `MulticolorGridBoard`, and `FractionalBoard` may be able to implement all four (although I am hesitant to replace grid-based boards with graph-based boards for performance reasons).

merowin:
How should the last move marker for Taengeuk Stones look?
I like this idea, maybe it could also work well here: https://forums.online-go.com/t/fractional-multicolour-go/36996/64
(though maybe with filled circles for markers instead of squares)

benjaminpjones:
Yeah that's a tough one. The linked solution looks mostly good, but does it work for gray stones?  Another solution might be something like:

```
if (colors are mostly light) {
    useBlackIndicator();
} else {
    useWhiteIndicator();
}
```

JonKo314:
[...]

PR #310, opened 2024-10-22
add score board in multicolor boards

merowin:
This adds score boards to multicolor boards. ~~Only unicolor score because I did not know how to best display a multicolor score marking (rectangle), but I believe for all our current variants, this is sufficient.~~ <- added multicolor score markings

Also I didn't implement the mapping from game state in all boards - I'm thinking we should probably try to reduce the number of board implementations, I'm just unsure how.

This is one step for #137

grafik

On graph boards it looks a bit funny, maybe due to rounding problems (?)

grafik


benjaminpjones (approved):
Very nice, thanks!

> On graph boards it looks a bit funny, maybe due to rounding problems (?)

TBH looks okay to me, but also I wonder if the shape (square) could be a source of some weirdness on non-grid boards.  We could potentially try out circles or other shapes.

PR #311, opened 2024-10-25
Add enforce-style-attribute lint rule

benjaminpjones:
Inspired by  #298, `scoped` seems to be best practice.  Let's enforce it everywhere.

### DIFF
--- a/packages/vue-client/.eslintrc.cjs
+++ b/packages/vue-client/.eslintrc.cjs
@@ -18,5 +18,6 @@ module.exports = {
         varsIgnorePattern: "^_",
       },
     ],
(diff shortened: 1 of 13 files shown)


merowin (approved):
Yeah, I strongly agree with this!

There are still ways to manipulate styles out of scope, particularly ":deep(selector)" seems useful.

There's also ":global" but I think we should use main.css for that.

benjaminpjones:
> There are still ways to manipulate styles out of scope, particularly ":deep(selector)" seems useful.

> There's also ":global" but I think we should use main.css for that.

Didn't know that! Glad we have an escape hatch, though I agree main.css seems to serve this purpose well.

PR #312, opened 2024-10-25
Remove Ko-detection in Thue-Morse

### LINKED ISSUE #200: [Thue-Morse] Refinements
Couple of ideas:

1.) Remove the check for (Super-) Ko, because the non-periodical move order rules out all cycles already
2.) Add visual indicator for the next moves that are coming up (e.g. display the next 6 or so move colors) <- this could also be scrollable

An idea (+ request) from a Thue-Morse Go enthusiast: Display the current move number in binary / hexadecimal / base 8

merowin:
closes #200

### DIFF
--- a/packages/shared/src/lib/ko_detector.ts
+++ b/packages/shared/src/lib/ko_detector.ts
@@ -1,4 +1,8 @@
-export class SuperKoDetector<T> {
+export interface KoDetector {
+  push(state: object): void;
+}
+
+export class SuperKoDetector<T extends object> implements KoDetector {
   seen = new Set<string>();
 
   push(state: T): void {
@@ -10,4 +14,8 @@ export class SuperKoDetector<T> {
 
 export class KoError extends Error {
   name = "KoError";
+}
+
+export class NoKoRuleDetector implements KoDetector {
+  push(_: object): void {}
 }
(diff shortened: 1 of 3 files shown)


benjaminpjones:
Change seems good, but I think doing this way may be a little hard to maintain (what if more gets added to `Baduk.postValidateMove()`?  do we need to update it again?)

Would it be possible to replace the `KoDetector` itself?  Something like:

```
class NullKoDetector {
    push() {
       // do nothing
    }
}
```

Then either pass it via config:

```
    // in Thue Morse constructor
    config.koDetection = "none"
    super(config);
```

Or by setting the `ko_detector` directly

```
   // In baduk.ts
   protected ko_detector: KoDetector;
   
   // in thue morse constructor
   ko_detector = NullKoDetector();
```

This would also help set us up for other forms of Ko detection (traditional Ko, PSK vs SSK distinction)

merowin:
> The idea seems good, but I think doing this way may be a little hard to maintain (what if we modify `Baduk.postValidateMove()`? How will we remember to update Thue-Morse accordingly?)

We could also consider adding a new function in Baduk just for the Ko check. Then inheriting classes could override such that it does nothing.

benjaminpjones:
> We could also consider adding a new function in Baduk just for the Ko check. Then inheriting classes could override such that it does nothing.

Yep! I think it's a variation on the same theme.  The only reason I lean toward replacing KoDetector objects instead of functions is because Ko detection is stateful, so we can abstract a bit more logic into the Detectors if we make it a class.

benjaminpjones (approved):
LGTM! thanks!

PR #313, opened 2024-10-25
Validate if there are follow-up moves in Keima variant

merowin:
Here are some change proposals to variant Keima based on the poll + discussion that we had here: [https://forums.online-go.com/t/collective-development-of-a-server-for-variants/43682/154](url)

- Validate and throw if there are no legal keima follow-up moves.
- Players can pass or play one keima, but they can't place a single stone and then pass.

This rules makes it so the variant basically can't be played on boards <= 2x2, but I only just realised that I should include a check for that.

### DIFF
--- a/packages/shared/src/variants/keima.ts
+++ b/packages/shared/src/variants/keima.ts
@@ -3,6 +3,7 @@ import { Grid } from "../lib/grid";
 import { getGroup } from "../lib/group_utils";
 import { Variant } from "../variant";
 import {
+  BadukConfig,
   BadukState,
   badukVariant,
   Color,
@@ -17,6 +18,13 @@ export interface KeimaState extends BadukState {
 
 export class Keima extends GridBaduk {
   private move_number = 0;
+
+  constructor(config: BadukConfig) {
+    super(config);
+    if (this.config.board.height < 3 && this.config.board.width < 3) {
+      throw new Error("Minimum board vor Keima is 3x2");
+    }
+  }
 
   playMove(player: number, move: string): void {
     super.playMove(player, move);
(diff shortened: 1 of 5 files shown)


packages/shared/src/variants/keima.ts
```diff
@@ -30,6 +42,24 @@ export class Keima extends GridBaduk {
...
         );
       }
+    } else if (
+      // check if there are any legal follow-up moves
+      getKeimaMoves(move, this.board).every((pos) => {
+        const boardCopy = this.board.map((x) => x);
+        // simulate a move
+        this.playMoveInternal(pos);
+        // this only checks if the follow-up move is suicidal, but it could also
+        // potentially be illegal due to ko.
```
benjaminpjones, line 52: Ah yeah, it's a little unfortunate we can't check Ko without disturbing the state of the KoDetector.  But I think we'd need to change the interface of both KoDetector and Baduk to check Future Kos easily.

benjaminpjones (approved):
Awesome, thanks!

> This rules makes it so the variant basically can't be played on boards <= 2x2, but I only just realised that I should include a check for that.

Ah yeah, seems like a good idea to add a check in the constructor. Though there's no harm in letting someone try it :smile:

merowin:
This pull request of mine has the same problem of maintainability as the other, because I'm calling functions of class Baduk that were previously private.

Other than this kind of "hack", I also considered implementing Keima such that the two moves that make up one keima are encoded in one move in the db. I think that would make the variant class simpler. The board component would need some additional changes.

benjaminpjones:
> I just realised this pull request has the same problem of maintainability as the other, because I'm calling functions of class Baduk that were previously private.

Hehe, this didn't bother me as much because the only assumption we made is that `playMoveInternal` will only change `board`.  IMO that's a reasonable assumption to rely on (but we should probably document it).



> I also considered implementing Keima such that the two moves that make up one keima are encoded in one move in the db. I think that would make the variant class simpler. The board component would need some additional changes.

Interesting idea.  I believe it would break SGF though :/

PR #315, opened 2024-10-30
fix bug where time control is undefined

merowin:
This aims to fix a time control bug that unfortunately I introduced with the game state changes.

The problem is when a move is played, the state sent does not include the time control object. As a consequence the clock does not display the times.

I've also made it so the game state map does not include the time control object any more ,for clarity and potentially saving memory space. I think even if users browse through the game history, its ok if we display the current times.

### DIFF
--- a/packages/vue-client/src/views/GameView.vue
+++ b/packages/vue-client/src/views/GameView.vue
@@ -13,6 +13,7 @@ import type {
   IPerPlayerTimeControlBase,
   IConfigWithTimeControl,
   GameStateResponse,
+  ITimeControlBase,
 } from "@ogfcommunity/variants-shared";
 import { computed, ref, watchEffect, type Ref } from "vue";
 import { socket } from "../requests";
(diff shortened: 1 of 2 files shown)


packages/vue-client/src/views/GameView.vue
```diff
@@ -40,11 +41,15 @@ const variantDescriptionShort = computed(() => getDescription(variant.value));
 const user = useCurrentUser();
 const playing_as = ref<undefined | number>(undefined);
 const displayed_round = computed(() => view_round.value ?? current_round.value);
+const time_control = ref<ITimeControlBase | null>(null);
 
 function setNewState(stateResponse: GameStateResponse): void {
+  const { timeControl: timeControl, ...state } = stateResponse;
+  if (timeControl) time_control.value = timeControl ?? null;
```
benjaminpjones, line 48: A little confused why we are checking `if (timeControl)` and also defining the `?? null` case.  Shouldn't one be sufficient?
```suggestion
  time_control.value = timeControl ?? null;
```

benjaminpjones (approved):
Nice, thanks for finding and fixing this bug!

packages/vue-client/src/views/GameView.vue
merowin: You are right, I think I made this a little bit hastily. But I'm not sure if we want to assign null, so I'm thinking
`if (timeControl) time_control.value = timeControl;`

PR #316, opened 2024-11-02
functionality to update rating and username in database

SameerDalal:
I created functionality to update user ratings as well as a simple function that is available if a username needs to be changed.

For now, the rating type is a number, but later it will be the type indicated by Glicko2. I wanted to add the Glicko2 library to the project in another commit for clarity.

### DIFF
--- a/packages/server/src/api.ts
+++ b/packages/server/src/api.ts
@@ -57,7 +57,7 @@ router.get("/game/:gameId/sgf", async (req, res) => {
     }
   } catch (e) {
     res.status(500).json({ error: e.message });
-  } 
+  }
 });
 
 router.get("/games/:gameId", async (req, res) => {
@@ -65,16 +65,15 @@ router.get("/games/:gameId", async (req, res) => {
     const game: GameResponse = await getGame(req.params.gameId);
 
     //updating player 0 rating to 600
-    const updateStatus = await updateUserRating(game.players[0].id, 600)
+    const updateStatus = await updateUserRating(game.players[0].id, 600);
     //accessing user from db to check if update was made
-    const userInfo = await getUserByUserID(game.players[0].id)
- 
+    const userInfo = await getUserByUserID(game.players[0].id);
+
     res.json({
       game,
       updateStatus,
       userInfo,
     });
-
   } catch (e) {
     res.status(500);
     res.json(e.message);
--- a/packages/server/src/api.ts
+++ b/packages/server/src/api.ts
@@ -15,8 +15,6 @@ import {
   createUserWithUsernameAndPassword,
   deleteUser,
   getUserByName,
-  updateUserRating,
-  getUser,
 } from "./users";
 import {
   getOnlyMove,
@@ -63,17 +61,7 @@ router.get("/game/:gameId/sgf", async (req, res) => {
 router.get("/games/:gameId", async (req, res) => {
   try {
     const game: GameResponse = await getGame(req.params.gameId);
-
-    //updating player 0 rating to 700
-    const updateStatus = await updateUserRating(game.players[0].id, 700);
-    //accessing user from db to check if update was made
-    const userInfo = await getUser(game.players[0].id);
-
-    res.json({
-      game,
-      updateStatus,
-      userInfo,
-    });
+    res.send(game);
   } catch (e) {
     res.status(500);
     res.json(e.message);
--- a/packages/server/src/api.ts
+++ b/packages/server/src/api.ts
@@ -15,6 +15,8 @@ import {
   createUserWithUsernameAndPassword,
   deleteUser,
   getUserByName,
+  updateUserRating,
+  getUserByUserID,
 } from "./users";
 import {
   getOnlyMove,
@@ -55,13 +57,24 @@ router.get("/game/:gameId/sgf", async (req, res) => {
     }
   } catch (e) {
     res.status(500).json({ error: e.message });
-  }
+  } 
 });
 
 router.get("/games/:gameId", async (req, res) => {
   try {
     const game: GameResponse = await getGame(req.params.gameId);
-    res.send(game);
+
+    //updating player 0 rating to 600
+    const updateStatus = await updateUserRating(game.players[0].id, 600)
+    //accessing user from db to check if update was made
+    const userInfo = await getUserByUserID(game.players[0].id)
+ 
+    res.json({
+      game,
+      updateStatus,
+      userInfo,
+    });
+
   } catch (e) {
     res.status(500);
     res.json(e.message);
--- a/packages/server/src/api.ts
+++ b/packages/server/src/api.ts
@@ -16,7 +16,7 @@ import {
   deleteUser,
   getUserByName,
   updateUserRating,
-  getUserByUserID,
+  getUser,
 } from "./users";
 import {
   getOnlyMove,
@@ -64,10 +64,10 @@ router.get("/games/:gameId", async (req, res) => {
   try {
     const game: GameResponse = await getGame(req.params.gameId);
 
-    //updating player 0 rating to 600
-    const updateStatus = await updateUserRating(game.players[0].id, 600);
+    //updating player 0 rating to 700
+    const updateStatus = await updateUserRating(game.players[0].id, 700);
     //accessing user from db to check if update was made
-    const userInfo = await getUserByUserID(game.players[0].id);
+    const userInfo = await getUser(game.players[0].id);
 
     res.json({
       game,
--- a/packages/server/src/users.ts
+++ b/packages/server/src/users.ts
@@ -49,25 +49,6 @@ export async function getUserByName(username: string): Promise<PersistentUser> {
   };
 }
 
-export async function getUserByUserID(
-  user_id: string,
-): Promise<PersistentUser> {
-  const db_user = (await usersCollection().findOne({
-    _id: { $eq: new ObjectId(user_id) },
-  })) as WithId<PersistentUser>;
-
-  if (!db_user) {
-    return undefined;
-  }
-
-  return {
-    username: db_user.username,
-    password_hash: db_user.password_hash,
-    login_type: db_user.login_type,
-    rating: db_user.rating,
-  };
-}
-
 export async function getUserBySessionId(
   session_id: string,
 ): Promise<GuestUser | undefined> {
@@ -198,6 +179,7 @@ function outwardFacingUser(
     id: db_user._id.toString(),
     login_type: db_user.login_type,
     ...(db_user.login_type === "persistent" && { username: db_user.username }),
+    rating: db_user.rating
   };
 }
 
--- a/packages/server/src/users.ts
+++ b/packages/server/src/users.ts
@@ -22,10 +22,13 @@ export async function updateUserRating(
   user_id: string,
   new_rating: number,
 ): Promise<void> {
-  await usersCollection().updateOne(
+  const update_result = await usersCollection().updateOne(
     { _id: new ObjectId(user_id) },
     { $set: { rating: new_rating } },
   );
+  if(update_result.matchedCount == 0){
+    throw new Error("User not found")
+  }
 }
 
 function usersCollection(): Collection<GuestUser | PersistentUser> {
--- a/packages/server/src/users.ts
+++ b/packages/server/src/users.ts
@@ -1,11 +1,12 @@
 import { getDb } from "./db";
 import { UserResponse } from "@ogfcommunity/variants-shared";
-import { Collection, WithId, ObjectId } from "mongodb";
+import { Collection, WithId, ObjectId, UpdateResult } from "mongodb";
 import { randomBytes, scrypt } from "node:crypto";
 
 export interface GuestUser extends UserResponse {
   token: string;
   login_type: "guest";
+  rating?: number
 }
 
 // Not currently used, but the plan is to use LocalStrategy from Password.js
@@ -14,25 +15,15 @@ export interface PersistentUser extends UserResponse {
   username: string;
   password_hash: string;
   login_type: "persistent";
-  rating?: number;
+  rating?: number
 }
 
-export async function updateUserRating(id: string, new_rating: number): Promise<void> {
-  if((await getUserBySessionId(id)).login_type == 'guest'){
-    throw new Error("Guest users cannot have ratings");
-  } else {
-    await usersCollection().updateOne(
-      { _id: new ObjectId(id) },
-      { $set: {rating: new_rating} }
-    );
-  }
-}
-
-export async function updateUsername(id: string, new_username: string): Promise<void> {
-  await usersCollection().updateOne(
-    { _id: new ObjectId(id) },
-    { $set: {username: new_username} }
+export async function updateUserRating(user_id: string, new_rating: number): Promise<UpdateResult<GuestUser | PersistentUser>> {
+  return usersCollection().updateOne(
+    { _id: new ObjectId(user_id) },
+    { $set: {rating: new_rating} }
   );
+
 }
 
 function usersCollection(): Collection<GuestUser | PersistentUser> {
@@ -56,11 +47,28 @@ export async function getUserByName(username: string): Promise<PersistentUser> {
   };
 }
 
+export async function getUserByUserID(user_id: string): Promise<PersistentUser> {
+  const db_user = (await usersCollection().findOne({
+    _id: { $eq: new ObjectId(user_id) },
+  })) as WithId<PersistentUser>;
+
+  if (!db_user) {
+    return undefined;
+  }
+
+  return {
+    username: db_user.username,
+    password_hash: db_user.password_hash,
+    login_type: db_user.login_type,
+    rating: db_user.rating
+  };
+}
+
 export async function getUserBySessionId(
-  id: string,
+  session_id: string,
 ): Promise<GuestUser | undefined> {
   const db_user = (await usersCollection().findOne({
-    token: { $eq: id },
+    token: { $eq: session_id },
   })) as WithId<GuestUser>;
 
   if (!db_user) {
--- a/packages/server/src/users.ts
+++ b/packages/server/src/users.ts
@@ -1,6 +1,6 @@
 import { getDb } from "./db";
 import { UserResponse } from "@ogfcommunity/variants-shared";
-import { Collection, WithId, ObjectId, UpdateResult } from "mongodb";
+import { Collection, WithId, ObjectId } from "mongodb";
 import { randomBytes, scrypt } from "node:crypto";
 
 export interface GuestUser extends UserResponse {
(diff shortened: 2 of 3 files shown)


SameerDalal:
If I run `yarn lint`  to fix the format changes, it modifies a lot of files that I have not touched.

In the build by Github it only shows one warn: `[warn] src/users.ts` but when running yarn lint:check locally I get warns from a lot of other files. Any reason why this is especially since I just pulled the latest code from the repo?

benjaminpjones:
> If I run yarn lint to fix the format changes, it modifies a lot of files that I have not touched.

This should not happen.  Are you sure it's actually modifying them?  When I run `yarn lint` it has a lot of logs but most of them are "src/some/file.ts (unchanged)

[yarn_lint_output.txt](https://github.com/user-attachments/files/17608776/yarn_lint_output.txt)

You can check which files have actually changed with `git status`.  If you are still having issues, I'd like to help - might help if you can run these again, then share the output.

```
git status
yarn lint
git status
```

SameerDalal:
> Are you sure it's actually modifying them? 

Git is showing that the files were modified, but I am not able to find a physical difference. The only file where I was able to find a difference was in the `users.ts` file, which I did add code to.
Here is the output that I am getting:

[terminal_output.txt](https://github.com/user-attachments/files/17608806/terminal_output.txt)

Initially there is nothing to commit but after running yarn lint there are numerous files that were changed

benjaminpjones (changes requested):
Thanks for working on this! Overall, I think it's good to figure out how we're going to store it in the database early, but it's not clear to me how this function is meant to be used.

Before merging, I'd like to be able to test this using some API.  A good test plan might be something like:

1) Finish a game with a user (or call whatever API we intend to call `updateUserRating()`)
2) Visit localhost:3000/api/games/<game_id>, and observe `players` contains rating == 0.0

BTW, if you need some way to observe/edit the local database without API, Mongo Compass is a nice tool

benjaminpjones:
> Git is showing that the files were modified, but I am not able to find a physical difference. The only file where I was able to find a difference was in the `users.ts` file, which I did add code to. Here is the output that I am getting:
> 
> [terminal_output.txt](https://github.com/user-attachments/files/17608806/terminal_output.txt)
> 
> Initially there is nothing to commit but after running yarn lint there are numerous files that were changed

Ah this is probably the dang [Windows/Unix newline issue](https://superuser.com/questions/374028/how-are-n-and-r-handled-differently-on-linux-and-windows).  I'm guessing it's changing all the `\n` to `\r\n`.  There is a way to make it work, but I am not so familiar with Windows myself.  @merowin do you have some advice?

packages/server/src/users.ts
```diff
@@ -14,6 +14,25 @@ export interface PersistentUser extends UserResponse {
   username: string;
   password_hash: string;
   login_type: "persistent";
+  rating?: number;
+}
+
+export async function updateUserRating(id: string, new_rating: number): Promise<void> {
+  if((await getUserBySessionId(id)).login_type == 'guest'){
+    throw new Error("Guest users cannot have ratings");
```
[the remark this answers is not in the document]
SameerDalal: You are right that enforcing this would add an additional query and slow things down. I just thought that since guests are temporary users, updating a rating for guests would be trivial.

packages/server/src/users.ts
```diff
@@ -14,6 +14,25 @@ export interface PersistentUser extends UserResponse {
...
+  } else {
+    await usersCollection().updateOne(
+      { _id: new ObjectId(id) },
+      { $set: {rating: new_rating} }
+    );
+  }
+}
+
+export async function updateUsername(id: string, new_username: string): Promise<void> {
```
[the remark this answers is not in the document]
SameerDalal: Got it!

packages/server/src/users.ts
```diff
@@ -14,6 +14,25 @@ export interface PersistentUser extends UserResponse {
...
   login_type: "persistent";
+  rating?: number;
+}
+
+export async function updateUserRating(id: string, new_rating: number): Promise<void> {
+  if((await getUserBySessionId(id)).login_type == 'guest'){
+    throw new Error("Guest users cannot have ratings");
+  } else {
+    await usersCollection().updateOne(
+      { _id: new ObjectId(id) },
```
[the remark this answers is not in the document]
SameerDalal: So, to get the User ID, would I use the `getUserBySessionId` function which would then pass to the `usersCollection().updateOne` function? For example:

``` TypeScript
const db_user_id = (await getUserBySessionId(id)).id;
  await usersCollection().updateOne(
    { id_: new ObjectId(db_user_id) },
    { $set: {rating: new_rating} }
  );
```

SameerDalal:
>Ah this is probably the dang [Windows/Unix newline issue](https://superuser.com/questions/374028/how-are-n-and-r-handled-differently-on-linux-and-windows)

I think this has something to do with the "linebreak-style" rule as I found a similar issue [here](https://stackoverflow.com/questions/39114446/how-can-i-write-a-eslint-rule-for-linebreak-style-changing-depending-on-windo) but have not been able to get any of their solutions working.

SameerDalal:
>I think it's good to figure out how we're going to store it in the database early, but it's not clear to me how this function is meant to be used.

Once I get Glicko2 into the project, I will have to use the `updateUserRating` function so that I can save the ratings for each user in the database. As you mentioned before, if I don't save it and the server restarts then the user rating information will be lost.

merowin:
> Ah this is probably the dang [Windows/Unix newline issue](https://superuser.com/questions/374028/how-are-n-and-r-handled-differently-on-linux-and-windows). I'm guessing it's changing all the `\n` to `\r\n` (or the other way around?). There is a way to make it work, but I am not so familiar with Windows myself. @merowin do you have some advice?

Unfortunately I don't have a solution either. I believe that I committed these "changes" once, and it didn't show as diff on GitHub, so I wonder if it actually changed anything.

packages/server/src/users.ts
```diff
@@ -14,6 +14,25 @@ export interface PersistentUser extends UserResponse {
   username: string;
   password_hash: string;
   login_type: "persistent";
+  rating?: number;
+}
+
+export async function updateUserRating(id: string, new_rating: number): Promise<void> {
```
merowin, line 20: Two thoughts:
1.) Currently this function doesn't allow the calling code to observe the result of this update function. I believe we can enable this by returning the Promise (see code example below).
2.) If `id` is the session id, then in my opinion the variable name should be more explicit, like `session_id`, otherwise I would expect that this is the user id.
On that note, why don't we pass the user id as input here? I think that might save us one data base query.

Example:
```
export async function updateUserRating(user_id: string, new_rating: number): Promise<UpdateResult<GuestUser | PersistentUser>> {
  return await usersCollection().updateOne(
    { _id: new ObjectId(user_id) },
    { $set: {rating: new_rating} }
  );
}

Then we can call this from outside and add error handling, e.g.:

`await updateUserRating(user_id, 1500).catch(error => console.error(error));`
```

benjaminpjones: Because it's marked "async", I believe it is returning a promise (but of course this will be more clear when we have the calling code)

benjaminpjones:
> I think this has something to do with the "linebreak-style" rule as I found a similar issue [here](https://stackoverflow.com/questions/39114446/how-can-i-write-a-eslint-rule-for-linebreak-style-changing-depending-on-windo) but have not been able to get any of their solutions working.

> Unfortunately I don't have a solution either. I believe that I committed these "changes" once, and it didn't show as diff on GitHub, so I wonder if it actually changed anything.

Okay, I wonder if [git is doing some processing](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings) as well.  In that case, I think it may be interfering with Prettier, but ultimately it will work okay.  So two options are:

1) try to commit with all the extra linefeed changes, and check git handles it okay.
2) manually unstage the extra changes, then commit.

---

>> I think it's good to figure out how we're going to store it in the database early, but it's not clear to me how this function is meant to be used.

> Once I get Glicko2 into the project, I will have to use the updateUserRating function so that I can save the ratings for each user in the database. As you mentioned before, if I don't save it and the server restarts then the user rating information will be lost.

Right, I understand this is intended to be used by the ratings code to save the ratings.  But I'd like to see the calling code before we merge.  This will answer questions like:

- Does the code actually work?
- Which ID should we use? Or rather, what user info do we have access to at the callsite?

packages/server/src/users.ts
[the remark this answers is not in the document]
benjaminpjones: Likely, we can forget about `getUserBySessionID`.  IIRC, that function only works for guest users anyway.

As @merowin suggested, can we make `user_id` the input to this function?

packages/server/src/users.ts
SameerDalal: In the example, if we use the UpdateResult Type directly from MongoDB then we should be able to just return `Promise<UpdateResult>` instead of `Promise<UpdateResult<GuestUser | PersistentUser>>` correct? 

Looking at the `updateOne` function, it doesn't return the updated user data, only the `UpdateResult` promise. So it may be misleading to specify that it is returning type GuestUser or PersistentUser.

SameerDalal:
>2. manually unstage the extra changes, then commit.

I can try this for my next commit, but it also means that for future commits, all the formatting changes would have to be done manually.

> But I'd like to see the calling code before we merge. This will answer questions like:

Got it. I will have some code up to test the function out.

packages/server/src/users.ts
merowin: > Looking at the updateOne function, it doesn't return the updated user data, only the UpdateResult promise. So it may be misleading to specify that it is returning type GuestUser or PersistentUser.

Hm as far as I can see the return type is `Promise<UpdateResult<TSchema>>`. In our case `usersCollection()` returns `Collection<GuestUser | PersistentUser>` i.e. `TSchema = GuestUser | PersistentUser`.

SameerDalal: Ah, I got it. Either way, it won't affect the return type so we can include the TSchema part like you said.

SameerDalal:
Using the "/games/:gameId" API, we can see that that the updateUserRating function is working correctly shown below. From the return value of the `usersCollection().updateOne()` function, the rating is being changed in the database as `modifiedCount = 1` and we can verify the value by getting the user information from the DB.

```
  "updateStatus": {
    "acknowledged": true,
    "modifiedCount": 1,
    "upsertedId": null,
    "upsertedCount": 0,
    "matchedCount": 1
  },
  "userInfo": {
    "id": "6727efd6a5411a8fea54eb70",
    "login_type": "persistent",
    "username": "sameerGuest",
    "rating": 700
  }
```

packages/server/src/users.ts
```diff
@@ -38,10 +50,10 @@ export async function getUserByName(username: string): Promise<PersistentUser> {
 }
 
 export async function getUserBySessionId(
-  id: string,
+  session_id: string,
```
benjaminpjones, line 53: Nice! (I had assumed it might be clear from the function name, but never hurts to be more explicit)

packages/server/src/api.ts
```diff
@@ -61,7 +63,17 @@ router.get("/game/:gameId/sgf", async (req, res) => {
 router.get("/games/:gameId", async (req, res) => {
   try {
     const game: GameResponse = await getGame(req.params.gameId);
-    res.send(game);
+
+    //updating player 0 rating to 700
+    const updateStatus = await updateUserRating(game.players[0].id, 700);
+    //accessing user from db to check if update was made
+    const userInfo = await getUser(game.players[0].id);
```
benjaminpjones, line 70: Why just player 0?  Just for testing? For future reference, it will be simple to make this happen for all players:

```
const userInfos = await Promise.all(game.players.map(player => getUser(player.id));
```

But also be aware that [`game.players` its elements are nullable](https://github.com/govariantsteam/govariants/blob/d3444cb8916145464c632929aac0378ba6a8f782/packages/shared/src/api_types.ts#L16), so there should be more checking like `if (players)` and `players.filter(player => player)`

packages/server/src/users.ts
```diff
@@ -14,6 +15,17 @@ export interface PersistentUser extends UserResponse {
   username: string;
   password_hash: string;
   login_type: "persistent";
+  rating?: number;
+}
+
+export async function updateUserRating(
+  user_id: string,
+  new_rating: number,
+): Promise<UpdateResult<GuestUser | PersistentUser>> {
```
benjaminpjones, line 24: I could be missing something, but according to [mongodb source](https://github.com/mongodb/node-mongodb-native/blob/00be68f/src/operations/update.ts#L33), `UpdateResult` is not generic - I'm not sure what specifying a template parameter does.

benjaminpjones (changes requested):
Thanks for testing this, and updating the function to handle user ID instead of the session token!

While I would have liked to see the call in a more realistic setting (i.e. we should be updating rating at game end or something, right?), I now have more confidence that `updateUserRating()` works as intended.

Let's revert the changes in `api.ts` for now, since it seems like those only serve testing purposes, then I can accept.

packages/server/src/users.ts
benjaminpjones: Not a huge deal, but I actually don't think we should return Mongo types like UpdateResult from functions in `users.ts`.  Basically I'd like to encapsulate such that if we ever tried to support a different DB we wouldn't need a bunch of API changes.  (also, I'm not even sure what to do with this result object as a caller)

packages/server/src/users.ts
SameerDalal: Yep! I will do the same for other functions that take id as a parameter

packages/server/src/api.ts
SameerDalal: > Just for testing?

Yes, I just wanted to make sure it worked, but I should have tested it for all players.

packages/server/src/users.ts
SameerDalal: I think I get what you mean. Hypothetically, if we keep it as is and use the return value for something then we later decide to switch to a new db, the way we use the return value would have to be changed. Is this what you mean? 

There's really no real purpose of the result object other than just making sure that values were updated. But now that we know that the function works, we can remove the return value.

packages/server/src/api.ts
SameerDalal: Also, the rating is only updated in the DB, so whenever we pull values from the DB to display to the user, would we also have to make sure to update the `game` property with the rating because the `game` property is what is passed to the user in the API?

packages/server/src/users.ts
merowin: It's ok to me if you remove the return value, I'm just throwing out thoughts because I believe it can be worth discussing things like this.

> Basically I'd like to encapsulate such that if we ever tried to support a different DB we wouldn't need a bunch of API changes.

Yeah it's a valid argument. My thought was / is that it's generally a good habit to have the most flexible / specific return type. You can always ignore the returned value.

> But now that we know that the function works, we can remove the return value.

Database operations / queries can have unexpected behaviour. It's possible that they fail unexpectedly, e.g. if something locks the entry you try to change. So in my opinion it's good to handle these things with care. It is a case where "it usually works", doesn't mean "it always works".

SameerDalal:
Also, I think I was not supposed to merge branch 'main' into this branch. I will try to see if I can undo it.

packages/server/src/users.ts
benjaminpjones: > Hypothetically, if we keep it as is and use the return value for something then we later decide to switch to a new db, the way we use the return value would have to be changed. Is this what you mean?

Correct!

> There's really no real purpose of the result object other than just making sure that values were updated. But now that we know that the function works, we can remove the return value.

There may be a purpose, but if there is, we should massage the value into something more understandable if we want to return something from this function.

> Database operations / queries can have unexpected behaviour. It's possible that they fail unexpectedly, e.g. if something locks the entry you try to change. So in my opinion it's good to handle these things with care. It is a case where "it usually works", doesn't mean "it always works".

Absolutely!  @merowin are there any states you think we need to bubble up to the user? I don't have a good understanding of this return object myself (upsertId, upsertCount etc.), and my hope was that all failures would lead to a rejected `Promise`.

packages/server/src/api.ts
benjaminpjones: > Also, the rating is only updated in the DB, so whenever we pull values from the DB to display to the user, would we also have to make sure to update the game property with the rating because the game property is what is passed to the user in the API?

I think that is something that can be decided.  But the important thing is this: if we store rating on the `game.players` object, it will not be synced to the user's current rating.  I think this is good, but just wanted to point it out.

packages/server/src/users.ts
merowin: > Absolutely! @merowin are there any states you think we need to bubble up to the user? I don't have a good understanding of this return object myself (upsertId, upsertCount etc.)

It seems that upsertId, upsertCount are only relevant when we pass the options { upsert: true } to the mutation function (e.g. updateOne, updateMany, maybe more). From [the docs](https://www.mongodbtutorial.org/mongodb-crud/mongodb-upsert/): 
```
Upsert is a combination of update and insert. 

Upsert performs two functions:

    Update data if there is a matching document.
    Insert a new document in case there is no document matches the query criteria.
```

I think the most interesting part about the UpdateResult might be `acknowledged`.

> all failures would lead to a rejected Promise.

This text in the node_modules files makes me think that we might just get the response `{ acknowledged: false }` in case of an error.

```
export declare interface UpdateResult<TSchema extends Document = Document> {
    /** Indicates whether this write result was acknowledged. If not, then all other members of this result will be undefined */
    acknowledged: boolean;
    [...]
}
```

benjaminpjones: Thanks for helping me understand the upsert stuff!

> we might just get the response { acknowledged: false } in case of an error.

If it is an indication of failure, one approach may be to turn it into a rejected Promise:

```
const result = await updateOne(...);
if (!result.acknowledged) {
    // throw in an async function is the same as rejected Promise
    throw new Error("Update not acknowledged");
}
```

Though based on [these docs](https://www.mongodb.com/docs/manual/reference/method/db.collection.updateOne/#returns), it seems possible acknowledgement is just telling us whether or not write concern is enabled.

> A boolean acknowledged as true if the operation ran with [write concern](https://www.mongodb.com/docs/manual/reference/glossary/#std-term-write-concern) or false if write concern was disabled

benjaminpjones (approved):

benjaminpjones:
Looks good, thanks again for working on ratings and also making the `id` parameters more descriptive.

packages/server/src/users.ts
SameerDalal: >Though based on [these docs](https://www.mongodb.com/docs/manual/reference/method/db.collection.updateOne/#returns), it seems possible acknowledgement is just telling us whether or not write concern is enabled.

After some testing, it seems that even if a faulty id is passed, `acknowledged`  is still true. I would say that a better indicator of failure would be `matchedCount` which tells us the number of documents that matched the filter. So we would have something like this:

```
if(result.matchedCount == 0){
    throw new Error("User not found")
  }
```

SameerDalal:
> Looks good, thanks again for working on ratings and also making the `id` parameters more descriptive.

Of course! But before we merge, I do want to add one last part to the code that manages the error handling.

PR #318, opened 2024-11-04
cleanup BadukWithAbstractBoard, move files, fix imports

merowin:
Sorry for the big diff. This is a bunch of cleanup of old stuff.

- Delete dead code: BadukBoardAbstract (was used before we had the board factory), IBadukBoard, StoneState, BadukWithAbstractBoardConfigForm
- Move board creation helpers into lib/abstractBoard/helper
- Move BadukWithAbstractBoard into variants/deprecated
- Export / Import fixes


packages/shared/src/index.ts
```diff
@@ -26,5 +26,6 @@ export * from "./lib/abstractBoard/intersection";
 export * from "./variants/s_fractional";
 export { count_binary_ones } from "./variants/thue_morse";
 export { sFractionalMoveColors } from "./variants/s_fractional";
+export * from "./variants/deprecated/badukWithAbstractBoard";
```
benjaminpjones, line 29: I'm a little surprised this needs to be exported now that everything had been moved out of here. Is this used in vue-client?

benjaminpjones (approved):
Thanks for the clean up!

packages/shared/src/index.ts
merowin: It's currently used in BadukBoardAbstract.vue <- this component only exists for mapping the config of legacy games of this variant to a Baduk config. So I'm thinking that we can add the mapping function as sanitizeConfig in the variant interface of BadukWithAbstractBoard, and then remove the above board component, as well as the export in index.ts.

Should I do it in this PR or a new one?

benjaminpjones: Separate PR would be fine! Mainly just curious how it gets used

PR #319, opened 2024-11-05
sanitize config of BadukWithAbstractBoard and remove its export and board component

merowin:
Follow-up for https://github.com/govariantsteam/govariants/pull/318 

More clean-up of the deprecated variant, tested locally


benjaminpjones (approved):
Super clean, thanks!

PR #320, opened 2024-11-05
add rules description for quantum

merowin:
This adds a rules description for quantum.

I've also noticed a (imo weird) css selector that changes the layout and font-weight of the rules description. I suspect this may be from the standard stylesheet, that we started out with. I think it's weird because it selects all elements in the dom, basically.


packages/vue-client/src/assets/base.css
```diff
@@ -93,6 +93,7 @@
   }
 }
 
+/* TODO: Check why this is here, can we remove this? */
```
benjaminpjones, line 96: 😯

Yeah idk what it's for, feel free to try removing it

benjaminpjones (approved):
Thanks for documenting quantum!

Cc: @SameerDalal

packages/vue-client/src/views/RulesView.vue
```diff
@@ -27,9 +27,7 @@ function toUpperCaseFirstLetter(string: string) {
 .rules-page {
   font-family: "helvetica", "arial", "sans-serif";
 }
-strong,
-h1,
-h2 {
+:deep(strong, h1, h2) {
```
benjaminpjones, line 30: I'm not familiar with :deep, is this a vue thing?

merowin: Yes I believe so: https://vuejs.org/api/sfc-css-features.html#deep-selectors

As I understand, it makes it so the css selector is applied "outside of scope" to this components child components.

The hint on this site is exactly the case for us:
`DOM content created with v-html are not affected by scoped styles, but you can still style them using deep selectors`

benjaminpjones: Cool, this seems useful.  Thanks for the pointer to the docs!

SameerDalal:
Thanks! Would it be fine if I send the .mp4 file? Not sure where I would upload the video, unless GoVariants has a Youtube channel.

SameerDalal:
Awesome! I can't upload it here since the file size is too big so I uploaded it to my drive [here](https://drive.google.com/file/d/1NC6p4yAt4Fe-2uf6jLgH8V-m4u1SYiP3/view?usp=sharing) where you can download it.

PR #322, opened 2024-11-09
Add a default board

benjaminpjones:
Based on discussion on #137, this is a proposal to replace Board .vue files with a simple JS function.  It's a larger PR, so to summarize:

- Introduce `DefaultBoard` (a selector for MulticolorGridBoard and MulticolorGraphBoard)
- Introduce `uiTransform` (a function that can update the config and state to fit the DefaultBoard interface)
- To validate, migrate `FreezeGo` to the `DefaultBoard` and test.

Testing: Freeze go and other variants seem to show up fine in the game list, game view, and demo view.

For example, here is the demo view:
Screenshot_20241108_225559


packages/shared/src/variants/freeze.ts
```diff
@@ -32,6 +40,34 @@ export class FreezeGo extends Baduk {
   override exportState(): FreezeGoState {
     return { ...super.exportState(), frozen: this.frozen };
   }
+
+  static uiTransform(config: NewBadukConfig, state: FreezeGoState) {
```
benjaminpjones, line 44: Note: we've added ~40 lines to this file, but removed over 100 lines by deleteing FreezeBoard.vue!

This is the benefit I am looking for (in addition to making boards more consistent)

packages/vue-client/src/components/boards/DefaultBoard.vue
```diff
@@ -0,0 +1,46 @@
...
+
+<script setup lang="ts">
+import MulticolorGridBoard from "./MulticolorGridBoard.vue";
+import { Coordinate, indexToMove } from "@ogfcommunity/variants-shared";
+import MulticolorGraphBoard from "./MulticolorGraphBoard.vue";
+import {
+  DefaultBoardConfig,
+  DefaultBoardState,
+  MulticolorStone,
+} from "@ogfcommunity/variants-shared/src/lib/board_types";
```
merowin, line 15: I've been trying to consistently import with "@ogfcommunity/variants-shared/" and then no path.

merowin (approved):
This is awesome! With this we can greatly simplify the board components, thank you! 😄

packages/vue-client/src/components/boards/DefaultBoard.vue
benjaminpjones: Good catch! There were quite a few of these dang auto imports. I will fix before landing.

benjaminpjones:
Changes after accept:

- Rebase (fixing minor merge conflict with #321 in GameListItem)
- Use standard import

packages/vue-client/src/components/boards/DefaultBoard.vue
benjaminpjones: related: https://github.com/govariantsteam/govariants/pull/323

PR #323, opened 2024-11-09
Add lint rule to prevent /src/ imports and fix instances.

benjaminpjones:
Now linter can catch the `@ogfcommunity/variants-shared/src/path/to/file` imports!


merowin (approved):
Neat, thanks for looking into this!

PR #326, opened 2024-11-11
pass round to playing table in demo

merowin:
This makes it so the upcoming moves display works in the demo board.

Remark about the displayed_round: Not all variant view components have this prop, and as a result I get this warning:

grafik

But in testing, it didn't seem to break anything (also in the game view component).

### DIFF
--- a/packages/vue-client/src/views/VariantDemoView.vue
+++ b/packages/vue-client/src/views/VariantDemoView.vue
@@ -44,6 +44,7 @@ const game = computed(() =>
     playing_as.value,
   ),
 );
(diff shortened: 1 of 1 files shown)


benjaminpjones (approved):
Oh yeah I noticed this when I added the getPlayingTable() function.  Thanks for adding it!

> Not all variant view components have this prop, and as a result I get this warning

That's unfortunate - mostly I think we should be able to ignore props we aren't using.  I don't mind the message, but if you find a way to suppress it, let me know!

PR #327, opened 2024-11-13
Migrate Baduk (and subvariants) to the DefaultBoard

benjaminpjones:
The main change is the addition of `uiTransform` to `baduk.ts` and deletion of the three  boards in vue-client/src/components/boards, but there are a bunch of knock-on effects because so many variants are built on top of this variant.

I'll try to comment on seemingly unrelated changes in this diff.

**Tested:** I spot checked affected variants

Screenshot_20241112_191002
Screenshot_20241112_190844


packages/shared/src/lib/board_types.ts
```diff
@@ -15,4 +15,7 @@ export type DefaultBoardConfig = {
 export type DefaultBoardState = {
   board: MulticolorStone[] | MulticolorStone[][];
   backgroundColor?: string;
+  // TODO: we should probably merge this into board
+  // score would be some field on MulticolorStone
+  score_board?: (string[] | null)[] | (string[] | null)[][];
```
benjaminpjones, line 20: I forgot to pipe this through in some way.  Also added the `TODO` since the "shapes" of `board` and `score_board` are similar

packages/shared/src/variants/baduk_utils.ts
```diff
@@ -61,3 +62,37 @@ export function mapToNewConfig<ConfigT extends LegacyBadukConfig>(
...
+  board: T[][],
+  f: (x: T) => S,
+  shape: "flatten-2d-to-1d",
+): S[];
+export function mapBoard<T, S>(
+  board: T[][],
+  f: (x: T) => S,
+  shape: BoardShape,
+): S[] | S[][];
+export function mapBoard<T, S>(
```
benjaminpjones, line 83: This function really helps with the differences in Grid/Graph.

packages/shared/src/variants/deprecated/badukWithAbstractBoard.ts
```diff
@@ -52,18 +52,12 @@ export function mapToNewBadukConfig(config: unknown): NewBadukConfig {
   return { komi: typedConfig.komi, board: mapToNewBoardConfig(typedConfig) };
 }
 
-export const badukWithAbstractBoardVariant: Variant<BadukConfig> = {
-  gameClass: Baduk,
+export const badukWithAbstractBoardVariant: Variant<
+  NewBadukConfig,
+  BadukState
+> = {
+  ...badukVariant,
```
benjaminpjones, line 59: This variant is functionally equivalent to Baduk, so I think copying in everything (except sanitizeConfig) makes sense.  I didn't think of a good way to test though :grimacing:

packages/shared/src/variants/freeze.ts
```diff
@@ -41,7 +41,10 @@ export class FreezeGo extends Baduk {
     return { ...super.exportState(), frozen: this.frozen };
   }
 
-  static uiTransform(config: NewBadukConfig, state: FreezeGoState) {
+  static uiTransform<ConfigT extends NewBadukConfig = NewBadukConfig>(
```
benjaminpjones, line 44: Had to change this because TypeScript wants types to match the parent class even for `static` methods. It's strange haha

packages/shared/src/variants/one_color.ts
```diff
@@ -13,7 +11,7 @@ export class OneColorGo extends Baduk {
   }
 }
 
-export const oneColorGoVariant: Variant<NewBadukConfig> = {
+export const oneColorGoVariant: typeof badukVariant = {
```
benjaminpjones, line 14: The alternative was to change to `Variant<NewBadukConfig, BadukState>`. I figured `typeof` was DRYer.

packages/vue-client/src/components/PlayingTable/ThueMorseTable.vue
```diff
@@ -1,12 +1,15 @@
...
+} from "@ogfcommunity/variants-shared";
 import MoveSequenceDisplay from "../MoveSequenceDisplay.vue";
 import { count_binary_ones } from "@ogfcommunity/variants-shared";
+import DefaultBoard from "../boards/DefaultBoard.vue";
 
 defineProps<{
-  gamestate: BadukState;
-  config: GridBadukConfig;
+  gamestate: DefaultBoardState;
+  config: NewGridBadukConfig;
```
benjaminpjones, line 12: This is okay because we have `sanitizeConfig`

packages/shared/src/variants/quantum.ts
```diff
@@ -310,4 +310,7 @@ export const quantumVariant: Variant<NewBadukConfig> = {
       komi: 7.5,
     };
   },
+  // quantum uses its own board, so we shouldn't use
+  // Baduk.uiTransform
+  uiTransform: undefined,
```
merowin, line 315: Maybe we can use the default board as child component of the quantum go board. But that's for the future anyway.

packages/vue-client/src/components/PlayingTable/ThueMorseTable.vue
merowin: Hm I think this should be `NewBadukConfig` though, or `DefaultBoardConfig`, since Thue-Morse can be played on graph boards too.

merowin (approved):
Very cool! 🎉

packages/shared/src/variants/quantum.ts
benjaminpjones: Oh yeah! Tbh I didn't even look at the impl of QuantumBoard when doing this, but migrating to DefaultBoard sounds like the right path!

packages/vue-client/src/components/PlayingTable/ThueMorseTable.vue
benjaminpjones: Oops you're right, just seeing this comment. I'll follow up

JonKo314:
Very cool indeed.

PR #328, opened 2024-11-14
Migrate QuantumBoard to DefaultBoard

benjaminpjones:
https://github.com/govariantsteam/govariants/pull/327#discussion_r1842799084

The major benefit is that we're able to remove some duplicate codepaths for Grid and Graph boards.  I did fudge types quite a bit, so I'm open to ideas on improvement, but overall, it works, and I think the reduction in code (~50 lines) is worth it.

Tested: both grid and graph boards

Screenshot_20241114_182904
Forgot to screnshot grid, but I promise I tested that too :)


JonKo314 (approved):

merowin (approved):
Looks good to me! 🙂

PR #330, opened 2024-11-15
migrate sfractional to default board

merowin:
This makes it so SFractional also uses the DefaultBoard component. Regarding the playing table, it works to have the uiTransform return an extended type (thanks for the suggestion @benjaminpjones !), and I think I found a relatively nice way to indicate the types.

relates to https://github.com/govariantsteam/govariants/issues/137


packages/shared/src/variants/s_fractional.ts
```diff
@@ -308,6 +313,32 @@ export class SFractional extends AbstractGame<
       secondary_colors: ["#0000FF", "#FF0000"],
     };
   }
+
+  static uiTransform(
+    config: SFractionalConfig,
+    gamestate: SFractionalState,
+  ): {
+    config: SFractionalConfig & DefaultBoardConfig;
```
benjaminpjones, line 321: I think you can just type this as `SFractionalConfig` and TypeScript will understand it satisfies `DefaultBoardConfig`.  But it's fine to keep it this way too.

packages/shared/src/variants/s_fractional.ts
```diff
@@ -308,6 +313,32 @@ export class SFractional extends AbstractGame<
...
+    gamestate: DefaultBoardState;
+  } {
+    const boardShape = isGridBadukConfig(config) ? "2d" : "flatten-2d-to-1d";
+
+    const stoneTransform = (colors: PlacementColors): MulticolorStone => {
+      return { colors: colors };
+    };
+    const scoreTransform = (color: string): string[] | null =>
+      color === "" ? null : [color];
```
benjaminpjones, line 330: Null checking not necessary here because `SFractionalState.score_board` is `string[][]` if it exists.

```suggestion
    const scoreTransform = (color: string): string[] => [color];
```

packages/vue-client/src/components/PlayingTable/SFractionalTable.vue
```diff
@@ -1,15 +1,16 @@
...
 } from "@ogfcommunity/variants-shared";
 import MoveSequenceDisplay from "../MoveSequenceDisplay.vue";
-import SFractionalBoardSelector from "../boards/SFractional/SFractionalBoardSelector.vue";
+import DefaultBoard from "../boards/DefaultBoard.vue";
 
 const props = defineProps<{
-  gamestate: SFractionalState;
-  config: SFractionalConfig;
+  gamestate: DefaultBoardState;
+  config: SFractionalConfig & DefaultBoardConfig; // return type of SFractional.uiTransform
```
benjaminpjones, line 13: Okay as-is, but TypeScript's `ReturnType` can make the comment something the compiler will understand.

https://www.typescriptlang.org/docs/handbook/utility-types.html#returntypetype

benjaminpjones (approved):
So much deleted code 😍

packages/vue-client/src/components/PlayingTable/SFractionalTable.vue
merowin: When I try `ReturnType<typeof sFractionalVariant.uiTransform>` I get the error
```
Type '((config: SFractionalConfig, state: SFractionalState) => { config: DefaultBoardConfig; gamestate: DefaultBoardState; }) | undefined' does not satisfy the constraint '(...args: any) => any'.
  Type 'undefined' is not assignable to type '(...args: any) => any'.ts-plugin(2344)
```
Is there a way to assert that it's not undefined?

packages/shared/src/variants/s_fractional.ts
benjaminpjones: Oh you're totally right. I misread it as saying color could be null.

packages/vue-client/src/components/PlayingTable/SFractionalTable.vue
benjaminpjones: Right, Variant.uiTransform is gullable. But you can do `ReturnType<typeof SFractional.uiTransform>` (the static method) which should be guaranteed to exist

packages/shared/src/variants/s_fractional.ts
benjaminpjones: Makes sense!

packages/vue-client/src/components/PlayingTable/SFractionalTable.vue
merowin: Oh yeah, I missed that, thank you! 😄 
With a structure like this, I feel like we are quite type safe.

packages/shared/src/variants/s_fractional.ts
```diff
@@ -327,11 +358,13 @@ export function sFractionalMoveColors(
...
+export const sFractionalVariant: Variant<SFractionalConfig, SFractionalState> =
+  {
+    gameClass: SFractional,
+    defaultConfig: SFractional.defaultConfig,
+    description:
+      "Stones have a secondary colour that also requires liberties to avoid capture.",
+    rulesDescription: sfractionalRulesDescription,
+    getPlayerColors: Baduk.getPlayerColors,
+    uiTransform: SFractional.uiTransform,
+  };
```
JonKo314, line 370: Why is this whole part changed? Was there some problem with indentation or is there now?

merowin: My auto formatter does this. It put the `{` bracked into the next line, because SFractionalState was added in the generics.

It also added indentation, I believe as an aftereffect of the above.

benjaminpjones: I think it's just the formatter adding whitespace (notice the `Variant<...>` part has been expanded)

JonKo314: Ah, maybe the line got too long, so Prettier put the curly bracket on the next line, then the indentation makes sense.

PR #331, opened 2024-11-15
pull out function examine group into utils

merowin:
This extracts the function examineGroup from SFractional into the group utils.  Maybe this function can serve as a generic helper function for variants like sfractional, similar to getGroup.

The returned array "adjacent" is not used currently, but I'm thinking this could be useful. I'd like to try and use it in future.


merowin:
Now that I think about it, I'll add a test case first. Back to WIP

benjaminpjones (approved):
This is awesome, and thanks for adding tests!

I'm also wondering if there are opportunities for synergies between the functions in this file.  Like seems like `getGroup` could potentially be based on this function.  Alternatively, this function could be a composition of `getGroup` and `getOuterBorder` (if those functions were generalized a bit)

Anyway, looks good as-is, just throwing out ideas for future improvements.

packages/shared/src/lib/group_utils.ts
```diff
@@ -74,3 +74,50 @@ export function getInnerBorder<K, V>(group: K[], graph: Fillable<K, V>) {
     group_on_graph.neighbors(index).some((index) => !group_on_graph.at(index)),
   );
 }
+
+export function examineGroup<K, V>(input: {
```
benjaminpjones, line 78: btw, love that this is generic now!

packages/shared/src/variants/s_fractional.ts
```diff
@@ -18,9 +18,10 @@ import {
   DefaultBoardState,
   MulticolorStone,
 } from "../lib/board_types";
+import { examineGroup } from "../lib/group_utils";
 
 declare type Color = string;
-declare type PlacementColors = [Color, Color] | [];
+declare type PlacementColors = Color[] & ([Color, Color] | []);
```
merowin, line 24: I changed this type to enable calling `PlacementColors.includes(string)`, but I just realised that this also allows developers to call `PlacementColors.push(string)`. So I guess the typing is not ideal here either way.

benjaminpjones: I think the idea behind tuple is that you should know what each member represents. So instead of:

```
color.includes("black");
```

You should be able to do something like:

```
// 0th element represents the "primary" color
color[0] === "black"

// or handling empty spaces
color.length && color[0] === black
```

merowin: The empty array in `[] | [string, string]` makes it so we can't call `includes(string)`, because it's interpreted as `never[]`. If we first check whether `array.length === 0`, then typescript understands, and allows the call to 'includes(string)'.

I'll reconsider my future type choices with this in mind, but I think it's ok to leave it like this for now.

benjaminpjones: Ohh I see now. I thought it was the other part causing issues!

I agree keeping includes makes sense even if type system is complaining.

PR #335, opened 2024-11-19
add quantum and sfractional to time handler map

merowin:
This adds Quantum and SFractional to the time handler map.

### DIFF
--- a/packages/server/src/time-control/time-handler-map.ts
+++ b/packages/server/src/time-control/time-handler-map.ts
@@ -23,4 +23,6 @@ export const timeControlHandlerMap: {
   keima: TimeHandlerSequentialMoves,
   "one color": TimeHandlerSequentialMoves,
   drift: TimeHandlerSequentialMoves,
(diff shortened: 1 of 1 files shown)


benjaminpjones:
Looks good!

Two comments, not directed at this PR:

Part of me wonders if this should be a required field of `Variant`, just so implementers get that reminder.  Wouldn't need to move all the code to shared/; it could look something like:

```
timeHandler: "sequential" | "parallel" | "none" | "custom";
```

Second, I notice on the main site, there is still an option to set a time config for these variants. We should not allow that for variants who don't specify a `timeHandler`.

benjaminpjones (approved):

merowin:
> Part of me wonders if this should be a required field of `Variant`, just so implementers get that reminder. Wouldn't need to move all the code to shared/; it could look something like:
> 
> ```
> timeHandler: "sequential" | "parallel" | "none" | "custom...?";
> ```

I like the idea! But I'd like to make a new PR for it.

PR #339, opened 2024-11-21
preselect seat in demo

### LINKED ISSUE #337: Play in Demo board without selecting seats
For all two-player variants, it would be nice if we can automatically select the next-to-play seat, so users don't need to click manually.

Maybe that could work something like this: When a move is played, if ´game.nextToPlay().length === 1´ then we automatically select the corresponding seat.

merowin:
closes #337


benjaminpjones (approved):
This will improve variant development immensely.  Thanks!

PR #340, opened 2024-11-22
Rules List

merowin:
This adds a new view that is linked to in the NavBar and has links to the rules page for variants with a (detailed) rules description.
Also I've added the demo board to the rules view as proposed in https://github.com/govariantsteam/govariants/issues/112
I wasn't able to make sure that the demo board fits, so I made its container scrollable on overflow.

### Rules List

grafik

grafik

grafik

### Rules Page

grafik

grafik

### DIFF
--- a/packages/vue-client/src/assets/base.css
+++ b/packages/vue-client/src/assets/base.css
@@ -93,7 +93,6 @@
   }
 }
 
-/* TODO: Check why this is here, can we remove this? */
 *,
 *::before,
 *::after {
--- a/packages/vue-client/src/utils/format-utils.ts
+++ b/packages/vue-client/src/utils/format-utils.ts
@@ -1,0 +1,7 @@
+export function toUpperCaseFirstLetter(string: string) {
+  if (string.length === 0) {
+    return "";
+  }
+
+  return string[0].toUpperCase() + string.slice(1);
+}
--- a/packages/vue-client/src/views/RulesListView.vue
+++ b/packages/vue-client/src/views/RulesListView.vue
@@ -1,0 +1,45 @@
+<script setup lang="ts">
+import { toUpperCaseFirstLetter } from "@/utils/format-utils";
+import { getVariantsWithRulesDescription } from "@ogfcommunity/variants-shared";
+import { RouterLink } from "vue-router";
+
+const variantsWithRulesDescription = getVariantsWithRulesDescription();
+</script>
+
+<template>
+  <main>
+    <h1>Rule Descriptions</h1>
+    <div class="rule-links-container">
+      <RouterLink
+        v-for="variant in variantsWithRulesDescription"
+        :key="variant"
+        class="rules-link"
+        v-bind:to="{ name: 'rules', params: { variant: variant } }"
+      >
+        {{ toUpperCaseFirstLetter(variant) }}
+      </RouterLink>
+    </div>
+  </main>
+</template>
+
+<style scoped>
+main {
+  padding: 0 2rem;
+}
+.rule-links-container {
+  display: flex;
+  flex-direction: row;
+  flex-flow: wrap;
+}
+.rules-link {
+  border: 1px solid var(--color-border);
+  margin: 15px;
+  text-align: center;
+  padding: 1rem 2rem;
+  font-size: large;
+  background-color: var(--color-background-soft);
+  &:hover {
+    background-color: var(--color-background-mute);
+  }
+}
+</style>
--- a/packages/vue-client/src/views/RulesListView.vue
+++ b/packages/vue-client/src/views/RulesListView.vue
@@ -14,7 +14,7 @@ const variantsWithRulesDescription = getVariantsWithRulesDescription();
         v-for="variant in variantsWithRulesDescription"
         :key="variant"
         class="rules-link"
-        v-bind:to="{ name: 'rules', params: { variant: variant } }"
+        :to="{ name: 'rules', params: { variant: variant } }"
       >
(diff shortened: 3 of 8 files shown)


benjaminpjones:
Very nice! These go great together, and I'm glad we're finally linking them from the main page.

I haven't had a chance to look at the code yet, but some visual thoughts:

- Text is a bit wide  IMO.
    - I searched around, and there is no one-size-fits-all approach.  However, most readability guidelines fall in the 45-90 characters range, whereas this is about 130 characters long.
    - Narrowing the text may also solve the space issue for the demo board as well. 
- On mobile, text could use some margin - I think it's like 3 px now, I'd recommend at least 8px
    - Compare Medium.com, which gives 24px on mobile
    - This might actually be some feedback for the website in general (config form also seems a bit cramped without margin), so we don't need to address in this PR

packages/vue-client/src/assets/base.css
```diff
@@ -93,7 +93,6 @@
   }
 }
 
-/* TODO: Check why this is here, can we remove this? */
```
benjaminpjones, line 96: Does this mean you figured it out?

packages/vue-client/src/utils/format-utils.ts
```diff
@@ -0,0 +1,7 @@
+export function toUpperCaseFirstLetter(string: string) {
```
benjaminpjones, line 1: Aside: We should really have some UI-friendly strings defined. I feel like the variant names are not designed to be shown to the user at all.

benjaminpjones (approved):
LGTM, thanks!

packages/vue-client/src/assets/base.css
merowin: I've tested around a bit, and realised 2 things:
1.) We can view these as changes to the "default styles" of css, and I think these are good.
2.) Components can still overwrite these styles if need be.

I added this TODO when I added some detailed rules, and the markdown is displayed differently than I expected. For the rules it might still be worth to overwrite this, but it's fine for now imo.

benjaminpjones: Makes sense, thanks for the explanation!

packages/vue-client/src/views/RulesListView.vue
```diff
@@ -0,0 +1,45 @@
...
+
+<template>
+  <main>
+    <h1>Rule Descriptions</h1>
+    <div class="rule-links-container">
+      <RouterLink
+        v-for="variant in variantsWithRulesDescription"
+        :key="variant"
+        class="rules-link"
+        v-bind:to="{ name: 'rules', params: { variant: variant } }"
```
JonKo314, line 17: FYI: Because of the inconsistency between `:key` and `v-bind:to` I want to point out, that `:` is a shorthand for `v-bind:`. See [Vue's v-bind directive](https://vuejs.org/api/built-in-directives.html#v-bind)

packages/vue-client/src/views/RulesView.vue
```diff
@@ -1,38 +1,46 @@
...
   </main>
 </template>
 
 <style scoped>
-.rules-page {
+main {
   font-family: "helvetica", "arial", "sans-serif";
 }
+
+.grid-page-layout {
```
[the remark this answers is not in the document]
benjaminpjones: Doesn't `<style scoped>` prevent the style from getting used in DemoView anyway (without deep selector)?

merowin:
>     * Text is a bit wide  IMO.
>       
>       * I searched around, and there is no one-size-fits-all approach.  However, most readability guidelines fall in the 45-90 characters range, whereas this is about 130 characters long.
>       * Narrowing the text may also solve the space issue for the demo board as well.
> 
>     * On mobile, text could use some margin - I think it's like 3 px now, I'd recommend at least 8px
>       
>       * Compare Medium.com, which gives 24px on mobile
>       * This might actually be some feedback for the website in general (config form also seems a bit cramped without margin), so we don't need to address in this PR

Thanks for the feedback! I've increased the padding on mobile to 12px (for all pages using grid-page-layout), and added some limits for the width of the text column.

packages/vue-client/src/views/RulesListView.vue
merowin: I've changed it, will try to be more consistent! 😄

packages/vue-client/src/views/RulesView.vue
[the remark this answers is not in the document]
merowin: Yeah this is what we renamed pageWrapper to. AFAIK with scoped it should only be applied to this component, and not its child components. see https://vue-loader.vuejs.org/guide/scoped-css.html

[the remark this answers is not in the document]
merowin: I'm not sure what the documentation means with `root node`. Does this mean the DOM tree?
But I tried it, and indeed the style is applied to both in this case. Good to know!
I've pushed a change that makes sure it's only applied to the rules page layout in this case.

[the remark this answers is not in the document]
JonKo314: DemoView has `<template><div class="grid-page-layout"> ... </div></template>`, I think the template is ignored, so the `<div class="grid-page-layout">` is the root node.

JonKo314 (approved):

packages/vue-client/src/views/RulesView.vue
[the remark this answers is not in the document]
benjaminpjones: Didn't know about that, thanks for explaining!

PR #342, opened 2024-11-24
fix liberty check for fractional go

JonKo314:
Wrong intersection indices were compared. E.g. the surrounded corner stone in the test was not checked, because it was `changedIntersection[7]` and `AbstractBaduk`'s `this.intersection[7]` was already checked as a part of the long black chain.

This fixes the [bug mentioned on OGS forums](https://forums.online-go.com/t/parallel-fractional-go-game-4/53434/100) of Game 4's round 16.

### DIFF
--- a/packages/shared/src/variants/fractional/fractional.test.ts
+++ b/packages/shared/src/variants/fractional/fractional.test.ts
@@ -6,7 +6,7 @@ import {
 } from "./fractional";
 
 class FractionalTestGame extends Fractional {
-  readonly firstCorner: FractionalIntersection | undefined;
+  readonly corners: FractionalIntersection[];
 
   constructor(config: Pick<FractionalConfig, "players">) {
     super({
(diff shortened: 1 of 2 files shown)


packages/shared/src/variants/fractional/fractional.test.ts
```diff
@@ -65,3 +60,27 @@ test("Surrounded merge stone", () => {
...
+      { primaryColor: "white", secondaryColor: "green" },
+    ],
+  });
+
+  expect(game.corners).toHaveLength(4);
+
+  game.playMove(0, game.idOf(game.corners[3]));
+  game.playMove(1, game.idOf(game.corners[3].neighbours[0]));
+
+  for (let i = 8; i >= 6; --i) {
```
benjaminpjones, line 77: Might be good to comment what is important about these moves. It's a little hard to parse out since a reader needs to do index calculation and know this is 9x9 board.  Or even a little ASCII art:

```
// . . . . . B B B B  // Additional chains on the board
// . . . . . . W W W  // have historically caused issues
// . ...5 lines... .
// W . . . . . . . .
//[B]W . . . . . . .  // White's last move is to capture black
```

benjaminpjones (approved):
Thanks for fixing!!  That doesn't look like an easy bug to find.

I don't have enough familiarity with `AbstractBaduk` to give a great review, but the presence of unit test gives me confidence.  Left a couple readability comments in there to make it very clear what the test is focused on.

PR #356, opened 2024-12-03
add link to element

merowin:
I added the link to Element (for inviting people) - though I'm not sure if this is the best URL for this purpose.

### DIFF
--- a/packages/vue-client/src/views/AboutView.vue
+++ b/packages/vue-client/src/views/AboutView.vue
@@ -15,6 +15,10 @@
             OGS Forums
           </a>
         </p>
+        <p>
+          <span>Or chat on </span>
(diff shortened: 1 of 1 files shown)


merowin:
> > though I'm not sure if this is the best URL for this purpose.
> 
> What does this mean? Is there another link we might consider? The invite link looks okay to me fwiw

I don't know if for example this would be better or not: `https://app.element.io/#/room/#govariants:matrix.org`
I don't know the nuanced differences.

benjaminpjones (approved):
Glad to have it!

JonKo314 (approved):
I don't really know, but I think matrix.to is there to have nicer links, so this should be fine.

PR #359, opened 2024-12-06
fix display bug in quantum

### LINKED ISSUE #358: Green colour does not get captured
In this game, when the white quantum stone gets captured, the green colour (quantum placement) does not get removed.

grafik
->
grafik

https://www.govariants.com/game/67527888998b9ba55f180008

merowin:
fixes #358 

Reason for the bug is that previously empty intersection was represented as undefined, now by empty array, and we used to call `colors?.push("green")` which used to only add green when the position is already occupied. So I added a function that does this for the new representation.


benjaminpjones (approved):
Thanks for fixing!

PR #360, opened 2024-12-07
complete quantum rules, improve tetris rules

merowin:
This adds a line in the quantum rules description for completeness, also I tried to improve the Tetris rules description.

### DIFF
--- a/packages/shared/src/templates/tetris_rules.ts
+++ b/packages/shared/src/templates/tetris_rules.ts
@@ -1,8 +1,8 @@
 export const tetrisRules = `
-Introduction
-------------
+Tetris Go is a variant of Go / Baduk with one additional rule:
 
-Tetris Go is a variant of Go / Baduk where in addition to the normal rules, moves that create a chain with a Tetris shape are not allowed.
-In other words, players can't create chains of size 4 (and this is how Tetris Go works on non rectangular boards).
-Note that players can still create chains of size 5 or more by connecting two chains.
+### Moves that create a chain with a Tetris shape are not allowed.
+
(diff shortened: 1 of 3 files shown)


packages/shared/src/templates/tetris_rules.ts
```diff
@@ -1,8 +1,8 @@
 export const tetrisRules = `
-Introduction
-------------
+Tetris Go is a variant of Go / Baduk with one additional rule:
 
-Tetris Go is a variant of Go / Baduk where in addition to the normal rules, moves that create a chain with a Tetris shape are not allowed.
-In other words, players can't create chains of size 4 (and this is how Tetris Go works on non rectangular boards).
-Note that players can still create chains of size 5 or more by connecting two chains.
+### Moves that create a chain with a Tetris shape are not allowed.
```
benjaminpjones, line 4: nit: Semantically, I don't think this is a "heading".  Might consider bold (markdown: `**bold text**`) or block quote (markdown `> quoted text`)

**Moves that create a chain with a Tetris shape are not allowed.**
> Moves that create a chain with a Tetris shape are not allowed.

If quote, we may need to add some global css, otherwise it just shows as a normal paragraph.

```
blockquote {
	border-left: 5px solid gray;
	margin: 1.5em 10px;
	padding: 0.5em 10px;
}
```

benjaminpjones (approved):

packages/shared/src/templates/tetris_rules.ts
merowin: I like the blockquote and added it, thank you for the suggestion!

Issue #336, opened 2024-11-19
Consistent Time Control Input Form


merowin:
The Input form for Time Control should only be displayed for variants with an entry in the `timeControlHandlerMap`.

Otherwise it looks like a bug when somebody creates a game, and the timer doesn't start. The absence of the input form could also help remember to add the variant to the map.

PR #361, opened 2024-12-07
Time handling in variant interface

merowin:
issue: #336 

I've made the changes suggested in https://github.com/govariantsteam/govariants/pull/335#issuecomment-2486948414
This adds a required field in the variant interface, but implementers can set it to "none" for starters.

On the server, I replaced the handler map by a function that uses the above.
The client does not displays the time config form when the above is set to "none".


packages/server/src/time-control/time-control.ts
```diff
@@ -59,6 +62,31 @@ export function ValidateTimeControlBase(
   );
 }
 
+/**
+ * Returns a new time handler instance suited for a given variant.
+ */
+export function GetTimeHandler(
```
benjaminpjones, line 68: nit - we should probably figure out what we want our naming conventions to be, but usually functions are lowerCamel (whereas UpperCamel is reserved for classes and types)

benjaminpjones (approved):
Awesome! Glad to have one fewer map around

packages/server/src/time-control/time-control.ts
merowin: Yeah I've been inconsistent, sorry about that. At work we have different technology stacks for frontend and backend, and the style guidelines also differ. So I'm used to e.g. write upper case in the backend.

Anyway, I renamed the function.

merowin: How do we want to write variables, by the way? I've seen Snake-Case, but I've also used Camel-Case.

benjaminpjones: I've been super inconsistent about this unfortunately. From a brief web search, it seems like lowerCamel might be more common for JS, generally.

- https://github.com/airbnb/javascript?tab=readme-ov-file#naming--camelCase
- https://google.github.io/styleguide/jsguide.html#naming-local-variable-names
- https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Writing_style_guide/Code_style_guide/JavaScript#variables

PR #368, opened 2025-01-01
Allow admins to delete users

benjaminpjones:
Basically what the title says.  Additionally, I pulled in the [sweetalert2](https://sweetalert2.github.io) library.  This gives us ability to launch more advanced modals than `window.alert()`, which was necessary for the additional verification in the User deletion flow.

0) Log in as admin
1) Click "Delete User"

2) Enter user's name as a safeguard
Screenshot_20241231_195147
3a) On failure, inform the admin
Screenshot_20241231_195215
3b) On success, inform the admin, then return home
Screenshot_20241231_195239

### DIFF
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -75,7 +75,9 @@ passport.deserializeUser<string>(function (id, callback) {
       if (user) {
         callback(null, user);
       } else {
-        callback(new Error(`No user with ID: ${id}`));
+        // This may be the case if a user has been deleted, but the user session persists.
+        // Sending null as the second argument invalidates the session.
+        callback(null, null);
       }
     })
     .catch((err) => {
--- a/packages/server/src/api.ts
+++ b/packages/server/src/api.ts
@@ -212,6 +212,23 @@ router.get("/users/:userId", async (req, res) => {
   }
 });
 
+router.delete("/users/:userId", async (req, res) => {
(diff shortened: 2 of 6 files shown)


packages/server/src/index.ts
```diff
@@ -75,7 +75,9 @@ passport.deserializeUser<string>(function (id, callback) {
       if (user) {
         callback(null, user);
       } else {
-        callback(new Error(`No user with ID: ${id}`));
+        // This may be the case if a user has been deleted, but the user session persists.
+        // Sending null as the second argument invalidates the session.
+        callback(null, null);
```
benjaminpjones, line 80: Without this change, the deleted user essentially gets "blacklisted" from the site: they just get the response body "No user with ID: <id>", until they clear their cache, which is not obvious to laypeople.

See also: https://github.com/jaredhanson/passport/issues/6

JonKo314 (approved):

benjaminpjones:
Going to merge since this is only exposed to Admins, but @merowin brought up a good point offline - we should remove deleted  players from their games as well.

I think this should be possible, and reasonably efficient using [`updateMany`](https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBCollection-updateMany/) along with [positional filtering](https://www.mongodb.com/docs/manual/reference/operator/update/positional-filtered/)

Until then - note to admins: only use this button if you are okay with their remnants lingering on games!

PR #369, opened 2025-01-02
fix distances of sierpinsky board

merowin:
I am happy to see that somebody played on the Sierpinsky board 😄 
I noticed that the stones overlap:

grafik

This PR fixes it:

grafik

### DIFF
--- a/packages/shared/src/lib/abstractBoard/helper/SierpinskyBoard.ts
+++ b/packages/shared/src/lib/abstractBoard/helper/SierpinskyBoard.ts
@@ -2,7 +2,7 @@ import { Vector2D } from "./types/Vector2D";
 import { Intersection } from "./types/intersection";
 
 export function createSierpinskyBoard(depth: number): Intersection[] {
-  const sideLength = Math.pow(2, depth + 1);
+  const radius = Math.pow(2, depth + 1) / Math.sqrt(3);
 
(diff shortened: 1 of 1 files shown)


packages/shared/src/lib/abstractBoard/helper/SierpinskyBoard.ts
```diff
@@ -2,15 +2,15 @@ import { Vector2D } from "./types/Vector2D";
 import { Intersection } from "./types/intersection";
 
 export function createSierpinskyBoard(depth: number): Intersection[] {
-  const sideLength = Math.pow(2, depth + 1);
```
benjaminpjones, line 5: Looks much better!

Just to make sure I understand the name change... is this the radius in question?

Screenshot 2025-01-02 at 9 01 48 PM

benjaminpjones (approved):
Looks nice!

I notice there's some commented code in this file that probably needs a similar fix.  Generally, I recommend deleting commented code due to the tendency to get out-of-date.

packages/shared/src/lib/abstractBoard/helper/SierpinskyBoard.ts
merowin: > Just to make sure I understand the name change... is this the radius in question?

Yes exactly 🙂

PR #371, opened 2025-01-06
Last move indicator in UI transform

merowin:
Issue: #137 
- circle annotation uses mix-blend-mode: difference
- enable the annotation in variants baduk, sfractional
- fix annotation in graph board

grafik

grafik

### DIFF
--- a/packages/shared/src/variants/baduk_utils.ts
+++ b/packages/shared/src/variants/baduk_utils.ts
@@ -65,7 +65,11 @@ export function mapToNewConfig<ConfigT extends LegacyBadukConfig>(
 }
 
 export type BoardShape = "1d" | "2d" | "flatten-2d-to-1d";
-export function mapBoard<T, S>(board: T[], f: (x: T) => S, shape: "1d"): S[];
+export function mapBoard<T, S>(
+  board: T[],
+  f: (x: T, idx: number | CoordinateLike) => S,
+  shape: "1d",
+): S[];
 export function mapBoard<T, S>(
   board: T[][],
   f: (x: T, idx: number | CoordinateLike) => S,
@@ -98,15 +102,22 @@ export function mapBoard<T, S>(
   }
 }
 
+/**
+ * Checks whether a string-encoded move is a stone placement at a certain place.
+ * @param move The string-encoded move to check
+ * @param idx An identifier for an intersection on the go board. If it is of type number,
+ * then we assume that we work with a graph board. In case of type CoordinateLike, we assume
+ * a rectangular board.
+ */
 export function equals_placement(
-  last_move: string,
+  move: string,
   idx: number | CoordinateLike,
 ): boolean {
-  if (["", "pass", "resign", "timeout"].includes(last_move)) {
+  if (["", "pass", "resign", "timeout"].includes(move)) {
     return false;
   }
   if (typeof idx === "number") {
-    return idx === Number(last_move);
+    return idx === Number(move);
   }
-  return Coordinate.fromSgfRepr(last_move).equals(idx);
+  return Coordinate.fromSgfRepr(move).equals(idx);
 }
--- a/packages/shared/src/variants/baduk_utils.ts
+++ b/packages/shared/src/variants/baduk_utils.ts
@@ -3,6 +3,7 @@ import {
   BoardPattern,
   GridBoardConfig,
 } from "../lib/abstractBoard/boardFactory";
+import { Coordinate, CoordinateLike } from "../lib/coordinate";
 import { Grid } from "../lib/grid";
 import { BadukConfig } from "./baduk";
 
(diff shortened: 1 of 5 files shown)


packages/shared/src/variants/baduk_utils.ts
```diff
@@ -67,22 +68,22 @@ export type BoardShape = "1d" | "2d" | "flatten-2d-to-1d";
 export function mapBoard<T, S>(board: T[], f: (x: T) => S, shape: "1d"): S[];
```
benjaminpjones, line 68: should this overload take idx too?

packages/shared/src/variants/baduk_utils.ts
```diff
@@ -67,22 +68,22 @@ export type BoardShape = "1d" | "2d" | "flatten-2d-to-1d";
...
 export function mapBoard<T, S>(
   board: T[][],
-  f: (x: T) => S,
+  f: (x: T, idx: number | CoordinateLike) => S,
```
benjaminpjones, line 71: We know what type `idx` is (`number` or `CoordinateLike`) based on `shape`.  I think it's like this:

```
mapBoard(board, f: (x: T, idx: CoordinateLike) => S, shape: "2d");
mapBoard(board, f: (x: T, idx: number) => S, shape: "1d");
mapBoard(board, f: (x: T, idx: number) => S, shape: "flatten-2d-to-1d");
```

If TypeScript complains, no worries, but I think this would be most accurate!

packages/vue-client/src/components/boards/MulticolorGraphBoard.vue
```diff
@@ -95,16 +95,16 @@ const viewBox = computed(() => {
...
-        )"
-        :key="index"
-        :cx="intersection.position.X"
-        :cy="intersection.position.Y"
-        :r="0.48"
-        :annotation="props.board.at(index)?.annotation!"
-      />
+      <template v-for="(intersection, index) in intersections">
+        <IntersectionAnnotation
+          v-if="props.board.at(index)?.annotation"
```
benjaminpjones, line 100: Does this change do anything?  IIUC it's equivalent to the `filter`

Just wondering in case the same code in `MulticolorGridBoard` needs to be fixed too.

packages/shared/src/variants/baduk_utils.ts
```diff
@@ -96,3 +97,16 @@ export function mapBoard<T, S>(
       return (board as T[][]).flat().map(f);
   }
 }
+
+export function equals_placement(
```
benjaminpjones, line 101: Maybe a doc comment here because:

a) This function is doing a bit more than just equality check
b) We have a few terms around for similar concepts: "index", "coordinate", "intersection".  Would be good to explain what we mean by "placement".

benjaminpjones (approved):
Great stuff, thanks! Just a few minor comments

packages/vue-client/src/components/boards/MulticolorGraphBoard.vue
merowin: Yes this is a bugfix. It's not equivalent because after filtering the array, the indices don't correspond to the indices of the original array any more. This made it so that `undefined` was passed as prop for `annotation`.

packages/shared/src/variants/baduk_utils.ts
merowin: Yes typescript complains, the reason being that the function type `(x:T, idx: number) => S` is not a special case of `(x: T, idx: number | CoordinateLike) => S`.

I believe it would work if I change the type thusly:
`f: ((x: T, idx: CoordinateLike) => S) | ((x: T, idx: number) => S)`
... and use type assertion in the implementation function.

packages/shared/src/variants/baduk_utils.ts
merowin: Yes I added it for consistency.

packages/shared/src/variants/baduk_utils.ts
merowin: Added a doc comment.

packages/vue-client/src/components/boards/MulticolorGraphBoard.vue
benjaminpjones: Oh makes sense! Grid.filter can preserve index but Array.filter cannot

packages/shared/src/variants/baduk_utils.ts
benjaminpjones: > I believe it would work if I change the type thusly:
f: ((x: T, idx: CoordinateLike) => S) | ((x: T, idx: number) => S)

Oh nice!

> ... and use type assertion in the implementation function.

IMO, a little type fudging in implementation can be worth it for more accurate interface (but nothing to spend too much time on)

merowin: I believe it would require changes where the functions are used, so I decided to land this regardless.

PR #372, opened 2025-01-24
more explicit types of board map functions

merowin:
Fixes technical debt that I left behind, see https://github.com/govariantsteam/govariants/pull/371#discussion_r1911865871

Makes the function types more explicit.

I would like to avoid the type assertions in the implementation function, but I don't know how.


benjaminpjones:
Thanks for circling back on this one! The updated signature looks good to me

benjaminpjones (approved):

PR #378, opened 2025-03-10
overwrite last move indicator in phantom

### LINKED ISSUE #377: Last move indicator is visible in Phantom.
Example: [https://www.govariants.com/game/67c634b494512450bc9de5a1](url)

I believe this bug was maybe introduced by adding the last move indicator in the Baduk variant.

merowin:
fixes #377


benjaminpjones (approved):
Good catch!

Issue #187, opened 2024-02-12
Board preview


merowin:
Not sure if we want to add this, but I just had this idea and wanted to write it down.

In the game creation form, we could display a preview of the board that will be used. This idea is mainly for variants with a new board pattern, so users can play with the parameters and find out what board it would be, without creating the game yet. It could be exclusive for certain variants (at the moment BadukWithAbstractBoard is ~the only variant with new board patterns~ and Fractional, and we could add this e.g. in their creation forms only)

Maybe a feature for the far future, as there are definitely more important things to work on.

benjaminpjones:
I think that would be great. And probably pretty easy - might extract some code from GameListItem and just skip the "for each move" logic

Another (heavier) approach - we might at some point want to have a graphical graph editor so that people can create boards without submitting code or editing JSON.  This could also serve as a way to visualize the board (even if the user is not editing)

benjaminpjones:
Not exactly this, but #249 does give a way to love update a board based on config changes. Same approach could be used for the game creation form.

PR #380, opened 2025-03-19
add board preview, expandable pocket component

merowin:
issue: #187 

Adds a board preview component, alongside a component "expandable pocket" which can be used to expand / collapse arbitrary content.

grafik

### DIFF
--- a/packages/vue-client/src/components/boards/BoardPreview.vue
+++ b/packages/vue-client/src/components/boards/BoardPreview.vue
@@ -1,0 +1,117 @@
+<script setup lang="ts">
+import {
+  BoardConfig,
+  createBoard,
+  createGraph,
+  Intersection,
+} from "@ogfcommunity/variants-shared";
+import { computed } from "vue";
+
+const props = defineProps<{
+  config: BoardConfig;
+}>();
+
+const isGrid = computed(() => props.config.type === "grid");
+
+// grid
+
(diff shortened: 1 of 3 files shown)


packages/vue-client/src/components/boards/BoardPreview.vue
```diff
@@ -0,0 +1,117 @@
+<script setup lang="ts">
```
merowin, line 1: There is some code duplicated from other board components here.

benjaminpjones: Wait, why copy the code?  This feels like something we can just use [`DefaultBoard`](https://github.com/govariantsteam/govariants/blob/020709a8145b2fa7c66e5327b1379433cf0de587/packages/vue-client/src/components/boards/DefaultBoard.vue) for

merowin: Hm good point, we could wrap it with `pointer-events: none` to prevent the hover effects, and get star points etc. for free.

merowin:
Ok, so I deleted the board preview component, and use the DefaultBoard instead. I made the game state prop optional, so we don't need to create a dummy state for the preview.
Also I changed the expandable pocket component so it looks nicer, including an expand / collapse animation.

grafik

PR #385, opened 2025-03-30
Update parallel.ts to exclude unstaged players from nextToMove()

benjaminpjones:
- `Indexes` -> `Indices`, move `getNullIndices()` into utils
    - Tbh I thought we had more than 2 parallel variants, hence the move to utils. 
- `parallel` to take the same `nextToMove` approach as `fractional` (see #381)

Tested:

- unit tests caught a lot!
- manually checked the parallel variant

### DIFF
--- a/packages/shared/src/variants/parallel.ts
+++ b/packages/shared/src/variants/parallel.ts
@@ -32,7 +32,10 @@ export class ParallelGo extends AbstractGame<
   private staged: MovesType = {};
   private last_round: MovesType = {};
   // Resignations and timeouts
-  private dropouts = new Set<number>();
+  private participants = Array.from(
+    { length: this.config.num_players },
+    (_, i) => i,
+  );
 
   constructor(config: ParallelGoConfig) {
     super(config);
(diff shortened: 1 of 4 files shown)


packages/shared/src/variants/parallel.ts
```diff
@@ -31,9 +31,8 @@ export class ParallelGo extends AbstractGame<
   private board: Grid<number[]>;
   private staged: MovesType = {};
   private last_round: MovesType = {};
-  private playerParticipation = this.initializeParticipation(
-    this.config.num_players,
-  );
+  // Resignations and timeouts
+  private dropouts = new Set<number>();
```
benjaminpjones, line 35: I noticed we aren't doing much with the playerParticipation data (i.e. round), so I decided to replace it with "dropouts" for easier access

benjaminpjones:
I had assumed it was used elsewhere, but if not, good idea!

Will push an update tonight

packages/shared/src/variants/parallel.ts
```diff
@@ -234,12 +229,10 @@ export class ParallelGo extends AbstractGame<
     });
   }
 
-  initializeParticipation(numPlayers: number): Participation[] {
-    const participation = new Array(numPlayers);
-    for (let i = 0; i < numPlayers; i++) {
-      participation[i] = { playerNr: i, dropOutAtRound: null };
-    }
-    return participation;
+  private participants() {
```
merowin, line 232: Seems like we use `participants` more than `dropouts`. Might be a little bit simpler if we turn it around - save participants in a variable, and remove a player when they drop out.

benjaminpjones: Good call! updated

PR #387, opened 2025-04-02
add the eqeqeq lint

benjaminpjones:
This feels like best practice (I had actually assumed this rule was already pulled in via `recommended`)


merowin (approved):
Thank you for adding this!

PR #389, opened 2025-04-02
Super tic tac go variant

goopletdev:
Added a new variant `SuperTicTacGo`, as well as a rules description. Included class `Nonant` to manage the 9 subsections of the goban.

Class SuperTicTacGo extends Baduk, and overrides `playMove()`, `postValidateMove()`, and `prepareForNextMove`.


benjaminpjones:
Code looks solid, and I tested it out manually - interesting variant! Nice work!

- `yarn lint` to fix signals
- Could you add a couple tests in variants/__tests__? It helps us to not break variants during refactors

packages/shared/src/variants/super_tic_tac_go.ts
```diff
@@ -0,0 +1,159 @@
...
+     * X and Y specify which nonant - 0-index, from top-left. 
+     * @param x 0 | 1 | 2
+     * @param y 0 | 1 | 2
+     */
+    constructor(
+        public readonly x: number,
+        public readonly y: number,
+    ) {
+        // there must bea better way 
+        // to specify range in typescript than this...
```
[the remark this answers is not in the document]
merowin: There is also Number.isInteger [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger](Url) and Number.isSafeInteger https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger that could be helpful.

packages/shared/src/variants/super_tic_tac_go.ts
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    
+    override postValidateMove(): void {
+        // Allow self-capture moves
+        // allow  ko
+    }
+
+    override prepareForNextMove(move: string): void {
+        if (move === 'pass') this.captures[this.next_to_play > 0 ? 0 : 1];
```
merowin, line 107: Is there maybe something missing here, like `this.captures[...]++` to increase it?

Variant Baduk is unfortunately inconsistent regarding the captures - we track them, but then `finalizeScore` uses area scoring. We should probably do something about that, but I'm not sure what.

[Unrelated comment] Scoring is generally a bit of a nightmare to implement 😅 We do it like in Tromp-Taylor rules, namely assuming that all stones on board are alive.

goopletdev: Ohhh right. Good catch, thanks 😅 I think for now I'd like to keep track of passing stones, even if they don't affect final score. 

Later, couldn't I override finalizeScore to just take captures into account? It doesn't have to be true territory scoring - for this particular variant, all that matters is that we're punishing passing moves. 

Although, it really needs quite a bit more playtesting before I'm even certain of that

packages/shared/src/variants/super_tic_tac_go.ts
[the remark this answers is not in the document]
goopletdev: I like the [].includes, i'll do that! Thanks.

packages/shared/src/variants/super_tic_tac_go.ts
benjaminpjones: I'm fine with this implementing pass stones, but I actually don't think the `captures` field is used for anything except Capture Go

> It doesn't have to be true territory scoring - for this particular variant, all that matters is that we're punishing passing moves.

Baduk implements area scoring, so passing (during dame) is punished.

packages/shared/src/variants/__tests__/super_tic_tac_go.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    [_, W, _, _, _, _, _, _, _],
+    [_, _, _, _, _, _, _, _, _],
+
+    [_, _, _, _, _, _, _, _, _],
+    [_, _, _, _, B, _, _, _, _],
+    [_, _, _, _, _, _, _, _, _],
+
+    [_, _, _, _, _, _, _, _, _],
+    [_, _, _, _, _, _, _, _, _],
+    [_, _, _, _, _, _, _, _, _],
```
goopletdev, line 40: I'd initially written this with two spaces in each line, which when combined with the two blank lines, neatly separated the 9x9 board visually into 9 Nonants. But alas, `yarn lint` seems to have undone that

packages/shared/src/variants/super_tic_tac_go.ts
goopletdev: > Baduk implements area scoring, so passing (during dame) is punished.

> Edit: oh wait, I think i understand now - there are some non-dame moves where one might want to pass, but should be prevented?
That was my thinking, based on the initial playtesting that we did, years ago. But I can't concretely remember whether we had even tried area scoring rather than territory, and i have a hunch that area coring may solve the problem altogether. needs more playtesting ^^

Incidentally, i imagine this variant would probably make territory scoring easy to implement, since the final board state generally has no dead stones on it.

packages/shared/src/variants/__tests__/super_tic_tac_go.test.ts
benjaminpjones: Ah yeah, [prettier](https://prettier.io) tends to discourage creative use of whitespace, and there's not a great way around it. I see what you were going for though!

packages/shared/src/variants/__tests__/super_tic_tac_go.test.ts
```diff
@@ -0,0 +1,42 @@
+import { SuperTicTacGo, Nonant } from "../super_tic_tac_go";
+import { Color } from "../baduk";
+
+const B = Color.BLACK,
+  _ = Color.EMPTY,
+  W = Color.WHITE;
+
+test("Play a game", () => {
```
benjaminpjones, line 8: Great test, thank you!

packages/shared/src/variants/__tests__/super_tic_tac_go.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  const game = new SuperTicTacGo({ width: 9, height: 9, komi: 0 });
+  expect(game.nextToPlay()).toEqual([0]);
+  expect(game.getLegalVertices().length).toEqual(81);
+  expect(() => game.getLegalNonant()).toThrow();
```
benjaminpjones, line 12: nit: for variant tests, I try not to test internals (meaning that I only test methods of `AbstractGame`)

In this case, we could probably remove all the `.getLegalNonant()` and `.getLegalVertices()` calls, as they are implicitly tested by `game.playMove()`.

Not only does this make future refactors easier, it also simplifies the test.  Then it's also possible to make these methods/classes private.

- https://en.wikipedia.org/wiki/Encapsulation_(computer_programming)
- https://softwareengineering.stackexchange.com/questions/100959/how-do-you-unit-test-private-methods

packages/shared/src/variants/super_tic_tac_go.ts
```diff
@@ -1,159 +1,179 @@
...
+      board: {
+        type: BoardPattern.Grid,
+        width: 9,
+        height: 9,
+      },
+    };
+  }
+
+  static sanitizeConfig(config: unknown): SuperTicTacGoConfig {
+    const sttgConfig = config as BadukConfig;
```
benjaminpjones, line 159: The reason `sanitizeConfig` takes `unknown` is that it is designed to sanitize legacy schemas (in other words, not guaranteed to be `BadukConfig` or whatever). Therefore, we usually do some proper type checking instead of type assertion (`as`).

In this case it might look like:

```
komi: 'komi' in config ? config.komi : 5.5;
```

https://www.allthingstypescript.dev/p/avoid-using-type-assertions-in-typescript

---

Separately, I think `sanitizeConfig` may be the wrong tool here.  I'd just `throw` in the constructor if you notice board is the wrong size/shape.  If a user enters a bad config, we will show the user a message like this:

Screenshot 2025-04-09 at 10 57 28 AM

packages/shared/src/variants/super_tic_tac_go.ts
```diff
@@ -1,159 +1,179 @@
...
+    if (this.last_move.length !== 2) {
+      throw new Error("all nonants are legal");
     }
+    const lastMove = super.decodeMove(this.last_move);
+    return Nonant.correspondingToCoordinate(lastMove);
+  }
 
-    protected hasOpenPoints(nonant: Nonant): boolean {
-        return this.getOpenVerticesInNonant(nonant).length > 0;
+  // mostly added this for easier testing
```
benjaminpjones, line 123: tbh, I think the test is fine without exposing this! I left another comment on the test about encapsulation

benjaminpjones (approved):
LGTM! Thanks for the lint and tests

packages/shared/src/variants/super_tic_tac_go.ts
goopletdev: oops, i got excited and hit 'merge' before seeing these comments >< i'm really not used to git workflow. do create a new pull request when i fix these things ?

Issue #346, opened 2024-11-24
Add CSFR middleware


benjaminpjones:
GitHub gave an alert saying we should do this. I believe the attack vector is that a user could visit a malicious site, which can apparently start sending API requests on behalf of the user.

PR #390, opened 2025-04-03
CSRF protection

merowin:
issue: #346 

Based on [https://medium.com/@brakdemir/csrf-prevention-on-node-js-express-without-csurf-da5d9e6272ad](url) using the Synchronizer Token Pattern

This is my attempt to add a system that protects us again Cross Site Request Forgery.

With CheckLogin, a random token is generated, stored in the users session, and sent in the response body. The client then sends this token in all requests header, and the server compares it to what was stored in session.


merowin:
I believe there is already an auth token that gets sent, but I suppose that is not enough to protect against CSRF for some reason.

I doubt that I implemented it s.t. there are no vulnerabilities, but maybe it's an okay start.

There are some open questions:
- What will happen with existing user sessions? I believe they will be unable to create games, and need to re-login. Can we do something about that?
- How to improve the types? (see TODOs)

merowin:
I tested that if the correct token is not provided, the api calls result in the correct error. And if it is provided, then they work (for user and guest user)

benjaminpjones:
I'm no expert, but looks about right to me.  In the future, we may want to consider outsourcing to an open source middleware like https://www.npmjs.com/package/csrf-sync, but I'm glad we have something now.

benjaminpjones (approved):

merowin:
> * What will happen with existing user sessions? I believe they will be unable to create games, and need to re-login. Can we do something about that?

I was wrong, as it is currently, a new token is generated whenever the CheckLogin request is used. The good news is, this way we don't need to worry about existing user sessions. The bad news is, I'm not sure if this is correct. The OWASP cheat sheet [https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html](url) specifies that the token can be generated per-request or per-session.

merowin:
> I'm no expert, but looks about right to me. In the future, we may want to consider outsourcing to an open source middleware like https://www.npmjs.com/package/csrf-sync, but I'm glad we have something now.

Yeah existing packages are surely better, but this way I learned something 😄

PR #391, opened 2025-04-03
shallow copy config to allow variants to mutate it

merowin:
Prevent an infinite loop that happens when variants mutate their config.

This is not a problem in GameView, because there we get the game state via server requests.

### DIFF
--- a/packages/vue-client/src/views/VariantDemoView.vue
+++ b/packages/vue-client/src/views/VariantDemoView.vue
@@ -46,6 +46,12 @@ const game = computed(() =>
 );
 const displayed_round = computed(() => view_round.value ?? moves.length);
 
+function cloneConfig(config: object): object {
+  // here we use this on a vue proxy object,
+  // and structuredClone doesn't work on those.
+  return JSON.parse(JSON.stringify(config));
+}
+
 function getGame(
   variant: string,
   config: object,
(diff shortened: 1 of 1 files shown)


benjaminpjones:
For my own understanding, why is shallow copy sufficient?

merowin:
> For my own understanding, why is shallow copy sufficient?

Now that you mention it, I don't think it is sufficient. Good catch!

benjaminpjones:
In that case, should be safe to JSON.stringify/JSON.loads approach, which is essentially what we do when we pass the config over the network.

merowin:
Is it ok if I use structuredCopy? I really dislike the `JSON.parse(JSON.stringify(...))` pattern.

benjaminpjones:
Usually, I'd agree, JSON is a primitive approach to cloning. However, in this case, `config` **must** be JSON-able, and this will be the closest approximation to the network path.

Also, I don't think Vue proxies are compatible with `structuredClone`? Anyway, there is a workaround, so that is an option: https://stackoverflow.com/questions/71075490/how-to-make-a-structuredclone-of-a-proxy-object

merowin:
Hm ok, I will try the Json way. (If structuredClone doesn't work, then JSON.stringify might also fail, I guess?)

Vue can be weird.

merowin:
The JSON clone works. I'm curious why structuredCopy doesn't work, but I'm glad that we have a working solution.

benjaminpjones (approved):
Thanks for fixing this!

PR #398, opened 2025-04-12
Doc comments on Variant.

### LINKED ISSUE #388: Doc comments on Variant
https://github.com/govariantsteam/govariants/blob/b96d1f52be658e8a1174b03b15949ece278543ff/packages/shared/src/variant.ts#L4

Most of these fields have very specific use-cases (but we haven't documented them). This might make it hard for new contributors to know when to use what.

Let's put doc comments on all the fields (especially the optional ones, which have fewer examples)

benjaminpjones:
Fixes #388 

This is one of the core classes used by variant implementers, but it's not always clear how these methods get used by the website.  Hopefully these docs clarify both intention and usage.

### DIFF
--- a/packages/shared/src/variant.ts
+++ b/packages/shared/src/variant.ts
@@ -29,7 +29,13 @@ export interface Variant<
    * completed) will remain accessible, but new games cannot be created.
    */
   deprecated?: boolean;
-  /** Used to determine the flow of the game.  In most cases, this is "sequential". */
+  /** Used to determine the flow of the game.
+   *
+   * Valid options:
+   * - "sequential" – Players take turns one after another.  This is the most common time control.
+   * - "parallel" – All players play a move in each round.
+   * - "none" – Time control is not supported.
+   */
   time_handling: "sequential" | "parallel" | "none";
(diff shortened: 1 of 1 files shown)


packages/shared/src/variant.ts
```diff
@@ -1,22 +1,51 @@
...
+   * Long description of the variant.  Supports markdown syntax.
+   * shown at /variants/<variant>/rules
+   */
   rulesDescription?: string;
+  /**
+   * Set this field to true to mark the variant as deprecated. Existing games (ongoing or
+   * completed) will remain accessible, but new games cannot be created.
+   */
   deprecated?: boolean;
+  /** Used to determine the flow of the game.  In most cases, this is "sequential". */
```
merowin, line 32: Maybe we could say that parallel means that there rounds where multiple players have a decision to make.

merowin (approved):
Looks good to me, thank you for adding this very important documentation!

packages/shared/src/variant.ts
benjaminpjones: Thanks for the feedback!  Added descriptions of each option.

PR #401, opened 2025-04-17
add voronoijs, shape click placeholder accordingly

merowin:
- Add typed package for computing Voronoi Diagram
- use voronoi diagram cells for intersection specific background colors (as in uiTransform).

### DIFF
--- a/packages/vue-client/src/components/boards/MulticolorGraphBoard.vue
+++ b/packages/vue-client/src/components/boards/MulticolorGraphBoard.vue
@@ -10,6 +10,7 @@ import {
   MulticolorStone,
 } from "@ogfcommunity/variants-shared";
 import ScoreMark from "./ScoreMark.vue";
+import { Voronoi, Site, Cell } from "voronoijs";
 
 const props = defineProps<{
   board?: (MulticolorStone | null)[];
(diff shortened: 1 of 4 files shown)


merowin:
I've reverted the changes on the click placeholders, and instead use the voronoi diagram only for the background color. I've also tested and made sure that the voronoi diagram is not computed when it is not being used at all.

packages/vue-client/src/components/boards/MulticolorGraphBoard.vue
```diff
@@ -22,6 +23,32 @@ const intersections = computed(() =>
...
+  return new Voronoi().compute(sites, {
+    xl: viewBox.value.minX - 0.5,
+    xr: viewBox.value.minX + viewBox.value.width + 0.5,
+    yt: viewBox.value.minY - 0.5,
+    yb: viewBox.value.minY + viewBox.value.height + 0.5,
+  });
+});
+
+function getPolygonPointsString(cell: Cell): string {
+  return cell.halfedges
```
merowin, line 43: The docs of voronoijs mention that it's possible that halfedges is empty array, which means that no cell could be calculated for this point. We may want to catch that case somehow, though for now I've decided to keep it simple.

benjaminpjones: I'm curious what scenario halfedges would be empty.. it must be rare, right?

benjaminpjones (approved):
Nice!

packages/vue-client/src/components/boards/MulticolorGraphBoard.vue
merowin: That is my assumption. Maybe when two points have the exact same coordinates, or something like that.

PR #408, opened 2025-04-27
Generalised pyramid counting

merowin:
# Proposed Changes
- Add graph utils including a function to calculate generalised pyramid weights
- Refactor variant Pyramid s.t. it supports graph boards (grid board uses the classical weights schema)
- Weights are calculated once on client and stored in config
- Fix a bug of voronoi diagram

# Note
- I noticed that the voronoi calculations sometimes crash, e.g. with sunflower board. I'll try to fix that.

grafik

### DIFF
--- a/packages/shared/src/variants/pyramid.ts
+++ b/packages/shared/src/variants/pyramid.ts
@@ -1,25 +1,55 @@
-import { Baduk, BadukState, Color, GridBaduk, gridBadukVariant } from "./baduk";
+import { Baduk, BadukBoard, BadukState, badukVariant, Color } from "./baduk";
 import { Grid } from "../lib/grid";
-import { GridBadukConfig, NewBadukConfig } from "./baduk_utils";
+import {
+  isGridBadukConfig,
+  isLegacyBadukConfig,
+  LegacyBadukConfig,
+  NewBadukConfig,
+  NewGridBadukConfig,
+} from "./baduk_utils";
 import { pyramidRuleDescription } from "../templates/pyramid_rules";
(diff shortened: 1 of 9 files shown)


packages/vue-client/src/components/boards/MulticolorGraphBoard.vue
```diff
@@ -31,12 +31,14 @@ const voronoiDiagram = computed(() => {
...
 
-  return new Voronoi().compute(sites, {
+  const diagram = new Voronoi().compute(sites, {
     xl: viewBox.value.minX - 0.5,
     xr: viewBox.value.minX + viewBox.value.width + 0.5,
     yt: viewBox.value.minY - 0.5,
     yb: viewBox.value.minY + viewBox.value.height + 0.5,
   });
+
+  return sites.map((site) => diagram.cells.find((cell) => cell.site === site)!);
```
merowin, line 41: I had to add this because `cells` is an unordered list.

packages/shared/src/variants/pyramid.ts
```diff
@@ -39,33 +68,46 @@ export class PyramidGo extends GridBaduk {
 
   private pointsForColor(c: Color) {
     if (this.score_board === undefined) return 0;
-    return this.score_board
-      .map((color, index) =>
-        color === c ? (this.weights.at(index) as number) : 0,
-      )
-      .reduce((x, y) => x + y, 0);
+    let points = c === Color.WHITE ? this.config.komi : 0;
```
merowin, line 71: this tixes pyramid not takint komi into account (the test case has 0 komi)

JonKo314 (approved):

Issue #226, opened 2024-03-30
Duplicate move on game


benjaminpjones:
This game got broken by a duplicate move: https://www.govariants.com/game/65cf8f1ee7d31b1d396fe3c7

benjaminpjones:
The race could look something like this:

    1) User POSTs game/xxx/move twice
        - Not sure how this happens, but server should be prepared for anything
    2) Call 1 pulls the game from the database
    3) Call 2 pulls the game from the database
    4) Call 1 pushes the move onto the end of the array (See [updateOne](https://github.com/govariantsteam/govariants/blob/7dbf9857e52442375a7faf9aa3909aab24c89c77/packages/server/src/games.ts#L171-L174))
    5) Call 2 pushes the move onto the end of the array (See [updateOne](https://github.com/govariantsteam/govariants/blob/7dbf9857e52442375a7faf9aa3909aab24c89c77/packages/server/src/games.ts#L171-L174))

A couple solutions I can think of:

- place a “pending” field on the game document using findAndModify, return early if the object is already being edited
    -  (This one seems really easy to screw up)
- Replace the array ($set) rather than just pushing ($push) the last move, so we always have a valid moves array
- We could add some self-healing, so if the db gets corrupted like this, we just remove the bad moves when the game is requested
    - This could be done in conjunction with fixing the race condition

benjaminpjones:
It looks like https://www.govariants.com/game/6776638b5c905eefea5d9d70 may be another example

merowin:
I've been looking into MongoDB Sessions and Transactions as a way to improve our DB consistency. From the docs:

https://www.mongodb.com/docs/manual/core/transactions-production-consideration/#in-progress-transactions-and-stale-reads

> In-progress Transactions and Write Conflicts
> 
> If a transaction is in progress and a write outside the transaction modifies a document that an operation in the transaction later tries to modify, the transaction aborts because of a write conflict.
> 
> If a transaction is in progress and has taken a lock to modify a document, when a write outside the transaction tries to modify the same document, the write waits until the transaction ends.

So I'm thinking that it might be sufficient to start a session and a transaction, and run all DB mutations with this session. For consistency we would need to pass the MongoDB session to all functions that modify the DB though (e.g. the ratings stuff).

merowin:
Another example: https://www.govariants.com/game/67e48a6ef7634fc792668d02

Beolach:
[...]

PR #411, opened 2025-05-01
admin edit game record

merowin:
issue: #226 
# Proposed changes
- add admin-only endpoints for fetching and updating a game record
- on client display textarea in GameView when in admin mode

# Notes
I realise that updating a game record in this way is very error prone. We should avoid it as much as possible.
Actually I built this before I had the idea for #410 . But this was still useful for testing #410 , so I thought I should at least put it up for debate.

### DIFF
--- a/packages/server/src/api.ts
+++ b/packages/server/src/api.ts
@@ -9,6 +9,7 @@ import {
   takeSeat,
   leaveSeat,
   getGameState,
+  gamesCollection,
 } from "./games";
 import {
   checkUsername,
@@ -29,6 +30,7 @@ import {
 } from "@ogfcommunity/variants-shared";
 import { io } from "./socket_io";
 import { checkCSRFToken, generateCSRFToken } from "./csrf_guard";
+import { ObjectId } from "mongodb";
 
 export const router = express.Router();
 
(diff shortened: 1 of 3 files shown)


packages/server/src/api.ts
```diff
@@ -314,3 +317,41 @@ router.get("/games/:gameId/state", async (req, res) => {
     res.json(e.message);
   }
 });
+
+router.get("/gameRecord/:gameId", checkCSRFToken, async (req, res) => {
+  if ((req.user as UserResponse).role !== "admin") {
```
merowin, line 322: We could extract this into a function like `checkCSRFToken`

benjaminpjones:
> I realise that updating a game record in this way is very error prone.

Yeah, I think we probably don't want to do this.  While I think this is reasonably scoped, admins already have ways to directly edit records.

And then, yeah I think something like repair (#410) is a nice approach with guard rails.

> But this was still useful for testing https://github.com/govariantsteam/govariants/pull/410

Have you tried [Compass](https://www.mongodb.com/products/tools/compass)?

merowin:
> Have you tried [Compass](https://www.mongodb.com/products/tools/compass)?

Oh thank you for the hint, no I haven't! I'll keep this in mind, though currently when I open the site, it gets stuck in a infinite reload loop 😅 Hopefully they'll fix it soon.

PR #413, opened 2025-05-04
Solve move race condition, add games schema

merowin:
This is a proposal for solving the race condition bug in #226 
We introduce a new field for game documents `lastMoveTime`, and rewrite the db update call s.t. it fails if this field changed in the meantime.
This has the following advantages:
- No usage of transaction (this would add some starting hurdle for devs, because they would need to make sure their local db instance is a replica set. Also makes the code more verbose, and as @benjaminpjones wrote, fundamentally transactions solve a different problem - atomicity for a set of db updates)
- In the scenario of two (almost) simultaneous calls to `playMove`, exactly one request throws an error.

Also adds a schema for games, because it was somehow necessary for this update call (plus makes things easier).

# Steps to test
add the following lines in `GameView` line 168
```
const test = {
  first: requests.post(`/games/${props.gameId}/move`, move).catch(alert),
  second: requests.post(`/games/${props.gameId}/move`, move).catch(alert),
};
```

### DIFF
--- a/packages/server/src/games.ts
+++ b/packages/server/src/games.ts
@@ -29,7 +29,6 @@ export type GameSchema = {
   config: object;
   players?: Array<User | undefined>;
   time_control?: ITimeControlBase;
-  lastMoveTime?: Date;
 };
 
 export function gamesCollection(): Collection<GameSchema> {
@@ -199,12 +198,9 @@ export async function handleMoveAndTime(
     .findOneAndUpdate(
       {
         _id: new ObjectId(game_id),
-        lastMoveTime: { $eq: game.lastMoveTime },
+        moves: { $size: game.moves.length },
       },
-      {
-        $push: { moves: moves },
-        $set: { time_control: timeControl, lastMoveTime: new Date() },
-      },
+      { $push: { moves: moves }, $set: { time_control: timeControl } },
     )
     .catch(console.log);
   if (!moveUpdateResult || !moveUpdateResult.lastErrorObject?.updatedExisting) {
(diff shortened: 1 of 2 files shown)


merowin:
@benjaminpjones no hurry on this. I'm requesting your review because it changes an important functionality.

packages/server/src/games.ts
```diff
@@ -186,12 +195,21 @@ export async function handleMoveAndTime(
     }
   }
 
-  gamesCollection()
-    .updateOne(
-      { _id: new ObjectId(game_id) },
-      { $push: { moves: moves }, $set: { time_control: timeControl } },
+  const moveUpdateResult = await gamesCollection()
```
merowin, line 198: This is also used in `timeout.ts`, where i wonder if we should change it s.t. the timeout move is always pushed on top, or not.

benjaminpjones: That's a good callout.

For  sequential, where the same user who plays is the one who times out, I think it's fine for the timeout to fail.

For the case of simultaneous moves, we'd want the timeout event to get pushed on.

benjaminpjones (approved):
In the same vein, `$size: game.moves.length` could probably be used as the staleness check.  It would help avoid an extra data field and also avoid any weird clock stuff (though I'm not that concerned about either).

---

Either way, I like this solution, and nice job finding a way to repro the issue!

packages/server/src/games.ts
merowin: Do you think it is ok to leave it as is for now? Currently I don't see many games of parallel variants played with time control at all (also, I assume the race condition with timeout would be pretty rare anyway).

PR #418, opened 2025-08-01
Docs baduk

merowin:


packages/shared/src/variants/baduk.ts
```diff
@@ -32,6 +32,14 @@ export enum Color {
 
 export type BadukConfig = LegacyBadukConfig | NewBadukConfig;
 
+/**
+ * Type representing the state of a Baduk game at one round. Contains all information used for displaying the (board-) state client side.
+ * @property board 2d-array containing the states (colors) of all board fields.
+ * @property next_to_play indicates which player can play a move next.
+ * @property captures maps index of a player to the number of their captures.
+ * @property last_move string encoding of the move played last round
```
benjaminpjones, line 40: might be good to specify the encoding (SGF coordinate)

packages/shared/src/variants/baduk.ts
```diff
@@ -175,6 +197,10 @@ export class Baduk extends AbstractGame<NewBadukConfig, BadukState> {
     });
   }
 
+  /**
+   * Called after placing stone and resolving captures. Validates if move is allowed by checking if it results in self-capture or a previously seen position.
```
benjaminpjones, line 201: These "Baduk lifecycle" methods especially benefit from the documentation, since many variants override them. 👍

Might be good to document what is expected of this function - specifically, that it should throw on an invalid move.

packages/shared/src/variants/baduk.ts
```diff
@@ -244,6 +273,10 @@ export class Baduk extends AbstractGame<NewBadukConfig, BadukState> {
     return this.sgf?.sgfContent ?? "non-rectangular";
   }
 
+  /**
+   * Initialises a Ko detector. Override this to change the Ko detection method.
```
benjaminpjones, line 277:
```suggestion
   * Override this to change the Ko detection method.
```

This is another case where I feel the name of the method pretty much said it already.

On the other hand, the notes about why one would override it are valuable.

benjaminpjones (approved):
Thank you for doing this!  I especially appreciate the documentation on the methods which are meant to be overridden.  I think that will be very helpful to variant authors.

packages/shared/src/variants/baduk.ts
benjaminpjones: Oh right! 😅 okay then fine as-is probably

PR #419, opened 2025-08-13
Remove baduk with abstract board

merowin:
- Remove legacy variant `BadukWithAbstractBoard`
- Remove backfill players array function
- yarn in migrations readme


benjaminpjones:
Thanks for cleaning up!

> Remove backfill players array function

Oh, nice - did these just get backfilled naturally?

merowin:
> Thanks for cleaning up!
> 
> > Remove backfill players array function
> 
> Oh, nice - did these just get backfilled naturally?

At one point, I queried the games where players array is empty using MongoDB Atlas. There were only a hand full, so I opened their game views, triggering the backfill.

I believe there was a single one left, because it was broken for a different reason (width value of 19.5 or something).

PR #434, opened 2026-01-03
Fix autoseating bug

### LINKED ISSUE #433: Auto-select seat on game reload
When I reload my game, I expect the only seat I am occupying to be selected, so I can play my move without confusion or interruption. But this doesn't happen. No seat is selected, and I must manually select my seat before playing.

Seems this was formerly done in pull-request #91 but has regressed?

benjaminpjones:
Fixes #433

Incidentally, we add our first Vue tests.  This is a subtle behavior that is easily missed in manual testing, and I want to prevent it from creeping back in.


benjaminpjones:
No problem, thanks for reviewing! And I have similar philosophy for ui tests - decided to remove all but one case (the one directly related to the bug)

PR #435, opened 2026-01-04
games endpoint returns game state and not all moves

merowin:
clean-up of endpoints that expose all moves of a game, instead send the game state.

### DIFF
--- a/packages/vue-client/package.json
+++ b/packages/vue-client/package.json
@@ -9,6 +9,7 @@
     "preview": "vite preview --port 4173",
     "build-only": "vite build",
     "type-check": "vue-tsc --noEmit -p tsconfig.vitest.json --composite false",
+    "test": "vitest",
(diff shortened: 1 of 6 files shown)


packages/vue-client/src/components/GameListItem.vue
```diff
@@ -1,29 +1,19 @@
...
-  };
-});
+const props = defineProps<{ game: GameInitialResponse }>();
 
 const transformedGameData = computed(() =>
-  uiTransform(props.game.variant, props.game.config, game.value.state),
+  uiTransform(
+    props.game.variant,
+    props.game.config,
+    props.game.stateResponse.state,
```
merowin: I feel like the type names are weird - instead of `GameResponse` (which is a much better name), now we send `GameInitialResponse`. Also this includes `stateResponse` which is an object that contains yet another property named `state` - quite confusing.

Better naming (and less nested structure) would be nice, but I didn't priortise it yet.

PR #436, opened 2026-01-07
Notifications

merowin:
Enable users to subscribe to certain game events and receive on-site notifications about them.

### Notification Types
- game end
indicates that the game is over, and includes the game result
- new round
indicates that a new round has started, and includes the round number
- my move
indicates that a new round has started where the user occupies a seat that can play a move this round
_Note: You can subscribe to other users games, but unless you occupy any seat, you won't receive a notification of this type._
- seat change
indicates that anybody has taken or left a seat

### Automatic deletion
To avoid spamming users with many notifications, we automatically delete old notifications when they are not interesting anymore. For example "Round 37 has started." then we delete "Round 36 has started." Similarly when a new round starts, then we delete existing "my move" notifications, since they refer to an old round.

### UI + how to test
1.) Subscribe to game notifications
In the game view click the bell icon in the upper right.

grafik

This brings up a dialog where you can check all notification types individually.

grafik

2.) Notifications view
The bell icon in the nav bar links to the notifications view, where your notifications are displayed grouped by game. The number of unread notifications is displayed as a badge.

The notifications view contains previews of the games which link to the corresponding game view. Users can also delete (bin icon) or mark their notifications as read (check icon), and configure their subscription (gear icon).

grafik

### DIFF
--- a/packages/server/src/games.ts
+++ b/packages/server/src/games.ts
@@ -239,12 +239,12 @@ export async function handleMoveAndTime(
       game.id,
       game_obj.round,
       userIdsOnThePlay,
-    ).catch((error) => console.error(error));
+    ).catch(console.error);
   }
 
   if (game_obj.phase === "gameover") {
     notifyOfGameEnd(game.subscriptions ?? {}, game.id, game_obj.result).catch(
-      (error) => console.error(error),
+      console.error,
     );
   }
 


packages/server/src/games.ts
```diff
@@ -22,14 +26,20 @@ import {
...
+  notifyOfNewRound,
+  notifyOfSeatChange,
+} from "./notifications/notifications";
 
 export type GameSchema = {
   variant: string;
   moves: MovesType[];
   config: object;
-  players?: Array<User | undefined>;
+  players: Array<User | undefined>;
```
benjaminpjones: is this a safe change now?  (looks good if so, but unrelated to notifs)

packages/server/src/notifications/notifications.ts
```diff
@@ -0,0 +1,194 @@
...
+      },
+    },
+  );
+}
+
+export async function notifyOfGameEnd(
+  subscriptions: GameSubscriptions,
+  gameId: string,
+  gameResult: string,
+): Promise<UpdateResult<UserNotifications>> {
```
benjaminpjones: IMO better not to leak the MongoDB return types.  The return value is not used at the callsite anyway, so something like `Promise<void>` or `Promise<bool>` would be sufficient and cleaner.

btw, what happens if db errors? Does this funciton throw?

packages/vue-client/src/main.ts
```diff
@@ -6,9 +6,20 @@ import router from "./router";
 
 import "./assets/main.css";
 
+// Vuetify
+import { createVuetify } from "vuetify";
+import * as components from "vuetify/components";
+import * as directives from "vuetify/directives";
+
+const vuetify = createVuetify({
```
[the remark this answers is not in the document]
merowin: However I didn't add the Vuetify styles, because they include global styles that messed up the rest of the UI.

packages/vue-client/src/views/GameView.vue
```diff
@@ -23,6 +24,12 @@ import PlayersToMove from "@/components/GameView/PlayersToMove.vue";
 import DownloadSGF from "@/components/GameView/DownloadSGF.vue";
 import { getPlayingTable } from "@/playing_table_map";
 import Swal from "sweetalert2";
+import SubscriptionDialog from "@/components/GameView/SubscriptionDialog.vue";
+import { FontAwesomeIcon } from "@fortawesome/vue-fontawesome";
+import { library } from "@fortawesome/fontawesome-svg-core";
+import { faBell } from "@fortawesome/free-solid-svg-icons";
+
+library.add(faBell);
```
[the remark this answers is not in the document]
merowin: I'll make a separate PR to clean up the library stuff (didn't think about it before to be honest)

packages/server/src/notifications/notifications.ts
merowin: I've removed the unused return types (changed to `Promise<void>`), and made it so the play move stuff does not await the notification functions, but exceptions get caught.

Question: What is a good way to handle such an exception? Currently I'm just calling `console.error(error)`, it would be nice if this shows up in the Heroku logs, but I don't know if it does.

merowin:
@benjaminpjones I changed the dialog to use sweetalert and removed Vuetify.

packages/server/src/api.ts
```diff
@@ -439,3 +438,114 @@ router.post("/admin/test-email", checkCSRFToken, async (req, res) => {
...
+
+      if (result.acknowledged) {
+        res.status(200);
+      } else {
+        res.status(500);
+      }
+      res.send({});
+    } catch (error) {
+      res.status(500).json({ error: error.message });
+      res.send({});
```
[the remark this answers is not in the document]
merowin: Should be fixed now with https://github.com/govariantsteam/govariants/pull/436/changes/8bec750e33ee2c560681fdaa61791c1c732e30bc

packages/server/src/notifications/notifications.ts
benjaminpjones: I think `console.log` does show up on heroku, though tbh I haven't checked

At some point it probably makes sense to use a "proper" logging solution like [Winston](https://www.npmjs.com/package/winston), but `console` API is fine for now.

packages/server/src/games.ts
```diff
@@ -233,16 +234,18 @@ export async function handleMoveAndTime(
...
       .map((index) => game.players[index]?.id)
       .filter((x) => !!x);
-    await notifyOfNewRound(
+    notifyOfNewRound(
       game.subscriptions ?? {},
       game.id,
       game_obj.round,
       userIdsOnThePlay,
-    );
+    ).catch((error) => console.error(error));
```
benjaminpjones: nit: can just pass the function

```
).catch(console.error);
```

PR #455, opened 2026-02-08
Quick fixes

merowin:
When logged-out I noticed an error in the console.
`Uncaught (in promise) TypeError: can't access property "user", e is null`
The reason is `CheckLogin` returns `null`, yet client tries to access `res` like an object. This makes it so `user` is set to `null` in this case.

I also noticed the bell icon looks like it's clickable, even though it's disabled when logged-out. This includes some style changes, to clear this up.

### DIFF
--- a/packages/vue-client/src/assets/main.css
+++ b/packages/vue-client/src/assets/main.css
@@ -136,12 +136,8 @@ blockquote {
     cursor: pointer;
   }
 
-  &:not(:hover) {
-    background-color: transparent;
-  }
-
+  &:not(:hover),
   &[disabled] {
-    filter: grayscale(0.9);
     background-color: transparent;
   }
 }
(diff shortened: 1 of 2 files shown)


benjaminpjones (approved):
thanks for the fixes!

disabled style looks good. were you able to check dark mode?  I find the contrast lacks for both this and the existing blue.

Image
Image

https://webaim.org/resources/contrastchecker/ is a helpful tool for this

merowin:
> thanks for the fixes!
> 
> disabled style looks good. were you able to check dark mode? I find the contrast lacks for both this and the existing blue.
> Image Image
> 
> https://webaim.org/resources/contrastchecker/ is a helpful tool for this

Oh that's really neat, thank you!

I've updated the color for better contrast (hopefully): 
grafik
grafik
grafik
grafik

PR #468, opened 2026-02-27
Variants improvements

merowin:
This is a follow-up to https://github.com/govariantsteam/govariants/pull/458 where I intend to add move preview implementations.
Also adds a utils function `map2d` which I found useful to simplify the move preview functions, and adds `getPlayerColors` to parallel variant.

### DIFF
--- a/packages/shared/src/variants/chess.ts
+++ b/packages/shared/src/variants/chess.ts
@@ -49,12 +49,24 @@ export class ChessGame extends AbstractGame<object, ChessState> {
   defaultConfig(): object {
     return {};
   }
+
+  static movePreview(
+    _config: object,
+    state: ChessState,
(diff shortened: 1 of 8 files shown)


packages/shared/src/variants/chess.ts
```diff
@@ -49,12 +49,24 @@ export class ChessGame extends AbstractGame<object, ChessState> {
...
     return {};
   }
+
+  static movePreview(
+    _config: object,
+    state: ChessState,
+    move: string,
+    _player: number,
+  ): ChessState {
+    const game = new Chess(state.fen);
```
benjaminpjones: this one is so clean - I wish Go could work like this!

benjaminpjones (approved):

PR #480, opened 2026-03-26
Align board styles across Fractional, Grid, and Graph boards

### LINKED ISSUE #137: Make boards more consistent

benjaminpjones:
## Summary
- Extract shared visual constants (`STONE_RADIUS`, `LINE_WIDTH`, `BOARD_COLOR`, `STAR_POINT_RADIUS`) into `board_constants.ts` so all three board components stay consistent
- Update FractionalBoard: stone radius 0.47→0.48, line weight 0.05→0.02, replace custom last-move circle with `IntersectionAnnotation`, add star points via `getHoshi`
- Remove dead `stroke="pink"` / `color="pink"` attributes from MulticolorGridBoard

Closes #137

## Test plan
just a visual check

Screenshot 2026-03-25 at 9 30 40 PM

### DIFF
--- a/packages/vue-client/src/components/boards/FractionalBoard.vue
+++ b/packages/vue-client/src/components/boards/FractionalBoard.vue
@@ -4,10 +4,18 @@ import {
   type FractionalConfig,
   type FractionalState,
   Intersection,
+  getHoshi,
 } from "@ogfcommunity/variants-shared";
 import { createGraph } from "@ogfcommunity/variants-shared";
 import { computed } from "vue";
 import TaegeukStone from "../TaegeukStone.vue";
+import IntersectionAnnotation from "../IntersectionAnnotation.vue";
+import {
+  STONE_RADIUS,
+  LINE_WIDTH,
+  BOARD_COLOR,
+  STAR_POINT_RADIUS,
+} from "./board_constants";
 
 const props = defineProps<{
   config: FractionalConfig;
@@ -43,6 +51,14 @@ const boardRect = computed(
   },
 );
 const board = computed(() => createGraph(intersections.value, null));
+
+const hoshi = computed(() => {
+  const b = props.config.board;
+  if (b.type === "grid") {
+    return getHoshi(b.width, b.height);
+  }
+  return [];
+});
 
 const viewBox = computed(() => {
   // viewBox is slightly larger than board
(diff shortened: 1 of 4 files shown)


JonKo314 (approved):

merowin (approved):
This is neat, thanks!

I've been thinking of ways how to use `DefaultBoard` for Fractional. I feel these changes reduce the need for that, but it might still be good for the future (?)

The main difference is that Fractional stores moves by index, i.e. Grid boards are treated like graph boards (so the size of grid boards is unlimited, whereas with sgf representation there is the limit of 52x52).
One way I've considered is that `FractionalBoard` could be a component wrapping `DefaultBoard` that converts the move representation before it is sent to server.
Another is to allow variants to use `MulticolorGraphBoard` also for grid boards. This would be a change in the `DefaultBoard`, and it could be represented by an optional flag in the board config.

PR #486, opened 2026-03-29
Validate socket subscriptions for seat-specific topics

### LINKED ISSUE #301: Server side validation of socket subscriptions
As part of our rework of the game state handling (#48) we introduced subscriptions that are used to make sure the different game states (as observed by different seats / an observer) are sent correctly. This is to protect hidden game information.

The api for such a subscription is currently lacking validation, i.e. clients could still subscribe even though they don't occupy this seat, and receive hidden information that way.

benjaminpjones:
## Summary

Socket connections previously had no user authentication — the server had no way to know who was connected. This meant any client could subscribe to any seat's topic (`game/{id}/{seat}`) and receive hidden game information. Additionally, the REST endpoint `GET /games/:gameId/state?seat=N` returned seat-specific hidden state with no auth check.

This PR fixes three attack vectors:

1. **Socket authentication**: Share Express session and passport with Socket.IO so socket connections carry user identity
2. **Subscription validation**: Before joining a seat-specific topic, verify the authenticated user actually occupies that seat
3. **REST endpoint protection**: `GET /games/:gameId/state?seat=N` now returns 403 if the user doesn't occupy the seat
4. **Socket eviction on seat change**: When a player sits or leaves, all sockets are evicted from the seat room so former occupants stop receiving hidden state

Observer topics (`game/{id}`) and seatless REST requests remain open to all — they only receive public state.

Fixes #301

## Test plan

- [x] `yarn build` — all packages build cleanly
- [x] `yarn test` — all 212 tests pass (6 new pure unit tests for subscription validation)
- [x] new tests cover: unauthenticated user, nonexistent game, wrong seat, empty seat, correct seat, and non-seat topics
- [x] wrote script to try to subscribe with the wrong seat. check server logs

```
  Rejected subscription to game/.../0: no authenticated user          ← unauth user                   
  Rejected subscription to game/.../0: user 69c89fe... does not occupy seat 0  ← wrong user
```

## Related

- Filed #487 for pre-existing `topics` array validation issue

### DIFF
--- a/packages/server/migrations/20260328000000-normalize-game-players.js
+++ b/packages/server/migrations/20260328000000-normalize-game-players.js
@@ -1,0 +1,104 @@
+module.exports = {
+  /**
+   * Convert players arrays from embedded user objects to user ID strings.
+   *
+   * Before: players: [{ id: "abc", username: "foo", ranking: {...} }, null]
+   * After:  players: ["abc", null]
+   *
+   * @param db {import('mongodb').Db}
+   * @param client {import('mongodb').MongoClient}
+   * @returns {Promise<void>}
+   */
+  async up(db, client) {
+    await client.withSession((session) =>
+      session.withTransaction(async () => {
+        const coll = db.collection("games");
+
+        // Find games where at least one player entry is an object (not yet migrated)
+        const games = await coll
+          .find({ "players.id": { $exists: true } })
+          .toArray();
+
+        if (games.length > 0) {
+          const ops = games.map((game) => ({
+            updateOne: {
+              filter: { _id: game._id },
+              update: {
+                $set: {
+                  players: game.players.map((p) =>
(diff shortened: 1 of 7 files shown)


packages/server/src/__tests__/subscription-validation.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    expect(mockGetGame).not.toHaveBeenCalled();
+  });
+
+  it("rejects seat topic when no user is authenticated", async () => {
+    const result = await validateSeatSubscription(
+      `game/${GAME_ID}/0`,
+      undefined,
+      mockGetGame,
+    );
+    expect(result).toBe("no authenticated user");
```
merowin: ideally, "magic strings" like this, that need to match with other files, should be extracted and imported from a central file.

PR #492, opened 2026-04-01
Fix O(n²) board construction for non-grid boards

benjaminpjones:
## Summary
- `convertIntersections` and `createGraph` in `boardFactory.ts` used `Array.indexOf()` to resolve neighbour references — O(n) per lookup, O(n²) overall
- Replaced with `Map<Intersection, number>` for O(1) lookups
- Affects all non-grid board types: sierpinski, polygonal, circular, trihexagonal, sunflower
- Sierpinski depth 10 (265K intersections): **31s → 200ms** in the perf test (156x speedup)

## Test plan
- [x] New perf test: sierpinski depth 8 board+graph creation under 500ms
- [x] `yarn build && yarn lint && yarn test` all pass
- [x] manually verify Sierpinsky board loads faster for sizes 3-8 

### DIFF
--- a/packages/shared/src/lib/__tests__/board-perf.test.ts
+++ b/packages/shared/src/lib/__tests__/board-perf.test.ts
@@ -1,0 +1,22 @@
+import {
+  BoardPattern,
+  createBoard,
+  createGraph,
+} from "../abstractBoard/boardFactory";
+import { Intersection } from "../abstractBoard/intersection";
+import { Color } from "../../variants/baduk";
+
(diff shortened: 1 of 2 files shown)


merowin (approved):
Nice, thanks!

About the size limit, I thought there was a client side limit in place for sierpinsky board (max = 6 on input type number), but seems like browsers don't care about that when you type a bigger number (?).
Would be good to add a limit on the server side too! Actually this ties in with the general problem that (I believe) the endpoint `create game` lacks validation.

PR #497, opened 2026-04-10
parallel staged moves default to empty in export state

### LINKED ISSUE #496: [Parallel] observer sees staged moves
In this game https://www.govariants.com/game/69bc593a64777f6e8551e2fd at move 40 as seat 0 I see 

Image

This is the correct board state with no staged moves.

But when no seat is selected, I see the staged move by seat 1

Image

merowin:
fixes #496

### DIFF
--- a/packages/shared/src/abstract_game.ts
+++ b/packages/shared/src/abstract_game.ts
@@ -34,7 +34,10 @@ export abstract class AbstractGame<
 
   abstract playMove(player: number, move: string): void;
 
-  /** Returns complete representation of the game at this point in time. */
+  /** Returns the game state from the perspective of a specific player or observer.
+   *
+   * @param player the 0-indexed seat number of a player, or undefined for observers.
+   */
   abstract exportState(player?: number): GameState;
 
   /**
--- a/packages/shared/src/variants/__tests__/parallel.test.ts
+++ b/packages/shared/src/variants/__tests__/parallel.test.ts
@@ -15,7 +15,7 @@ test("Halfway through the round, moves are staged, but not placed.", () => {
       [[], [], [], []],
       [[], [], [], []],
     ],
-    staged: { 0: "ba", 1: "ca" },
+    staged: {},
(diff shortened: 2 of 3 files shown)


merowin:
@benjaminpjones requesting your review, as the author of variant Parallel.

//Edit: Just seen that a test case fails. That makes me a bit unsure. But I think staged moves should be secret until the round transition.

benjaminpjones:
Surely moves should not be revealed to observers, so I think your fix is correct.

one thing to check before we land is whether our convention broke somehow. I believe there was (at some point) a way to differentiate observer (guest) vs. observer (omniscient - for like admin or post-game review).

I can take a look tonight and see if theres a story here. Likely your fix is correct and we can just update the test.

benjaminpjones:
Okay, I couldn't find evidence that `undefined` was supposed to be "omniscient", though apparently I did it this was for Phantom Go too.  I think the right move is to update the test and land this PR.

---

Claude's audit:

Variant | undefined behavior | Correct?
-- | -- | --
phantom.ts:8-10 | Returns full board (observer sees everything) | Leaks info vs. players — spectators see more than the players themselves
lighthouse.ts:50-51 | Maps to null, then exportFor(null, …) returns restricted intersection of both players' views | Explicit anti-cheat design
one_color.ts:4 | Ignores the parameter entirely; obfuscates for everyone | Consistent
parallel.ts:49-53 | Returns all staged moves (observers see every pending move) | Leaks pending moves to spectators
fractional.ts:108-119 | player != null check, so undefined yields no staged move | Safe
rengo.ts:62-67 | if (!player) — bug: player === 0 also takes the undefined branch, so player 0 gets the observer view instead of their team view |  

So some follow-ups:

a) document `exportState()` more clearly
b) fix bugs in `phantom` (#500) and `rengo` (#499)
c) reveal hidden info post-game (and for admin?)

benjaminpjones (approved):

PR #498, opened 2026-04-10
add game list store

### LINKED ISSUE #382: Preserve state of the games list
### Motivation
While browsing games on the site, I often find myself clicking into a game and later go back to Home. Then the games list is back to the start, but I would like to look at the next game(s). It would be great if the games list remembers its state in such a scenario.

### How
One possibility is to store the filter and pagination settings in local storage and then reload. But I think this is not necessary. It would suffice to extract the state of the games list component into an object that persists when the games list component is closed.

merowin:
closes #382

adds pinia store for game list state, so its state is (mostly) preserved through navigating to different parts of the app (e.g. when navigating game list -> game page -> back to game list, we are still on the same list page)

However this does not store the settings persistently over multiple sessions/devices as was discussed in the issue. I'd like to leave that for a future issue.

### DIFF
--- a/packages/vue-client/src/components/GameList.vue
+++ b/packages/vue-client/src/components/GameList.vue
@@ -1,46 +1,21 @@
 <script setup lang="ts">
-import { ref, computed, type Ref } from "vue";
-import { useFetch } from "@vueuse/core";
-import {
-  type GameErrorResponse,
-  type GameInitialResponse,
-  type GamesFilter,
-  gamesFilterToUrlParams,
-  isErrorResult,
-} from "@ogfcommunity/variants-shared";
+import { isErrorResult } from "@ogfcommunity/variants-shared";
 import GameListItem from "@/components/GameListItem.vue";
 import GameListItemFallback from "@/components/GameListItemFallback.vue";
 import GamesFilterForm from "@/components/GamesFilterForm.vue";
 import VErrorBoundary from "vue-error-boundary";
+import { gameListStore } from "@/stores/game-list-store";
+import { storeToRefs } from "pinia";
 
 const countOptions = [10, 15, 25, 50];
-const count = ref(countOptions[0]);
-const offset = ref(0);
-const filter: Ref<GamesFilter> = ref({});
-const url = computed(
-  () =>
-    `/api/games?count=${count.value ?? 0}&offset=${offset.value ?? 0}` +
-    gamesFilterToUrlParams(filter.value),
-);
-const { data: games } = await useFetch(url, { refetch: true })
-  .get()
-  .json<(GameInitialResponse | GameErrorResponse)[]>();
-const first = () => {
-  offset.value = 0;
-};
-const previous = () => {
-  offset.value = Math.max(0, offset.value - count.value);
-};
-const next = () => {
-  offset.value += count.value;
-};
-function setFilter(gamesFilter: GamesFilter) {
-  filter.value = gamesFilter;
-}
+
+const store = gameListStore();
+store.load();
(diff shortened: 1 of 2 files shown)


merowin:
@JonKo314 I'm requesting your review here as the original author of the game list pagination.

packages/vue-client/src/components/GameList.vue
```diff
@@ -1,46 +1,21 @@
...
-};
-const next = () => {
-  offset.value += count.value;
-};
-function setFilter(gamesFilter: GamesFilter) {
-  filter.value = gamesFilter;
-}
+
+const store = gameListStore();
+store.load();
```
benjaminpjones, line 13: Small downside of this refactor: the old `useFetch(url, { refetch: true })` automatically re-fetched whenever the computed URL changed, so adding a new filter field "just worked." Now every action has to remember to await `this.load()` at the end.

benjaminpjones (commented):
Curious why Pinia was chosen here over alternatives such as:

- URL params (discussed on issue, makes pages linkable, works even if you *leave* the site)
- `<KeepAlive>` (one-liner, preserves all component state)
- localstorage (persists across sessions - though you mentioned wanting to defer this feature)

The store is only consumed by GameList.vue, so we're not really getting the cross-component sharing benefit that a singleton store offers.

merowin:
> * URL params (discussed on issue, makes pages linkable, works even if you _leave_ the site)

I wasn't sure how that works when user input changes. Do I need to trigger a router navigation? (If so, I think that's a downside imo)
But tbh I also don't see much advantage. What would be an example scenario where it makes sense to link to a certain page? Let's keep in mind that [the games displayed for a certain page] change as more games are added (at least with the current implementation of the pagination. I know that e.g. Postgres offers a different along the lines of "first: 10, after: identifier") which is a particularily stable system)

> * `<KeepAlive>` (one-liner, preserves all component state)

To be honest I didn't know about this. Would it re-fetch the games in the scenario [user creates game, then navigates back to list]?

> * localstorage (persists across sessions - though you mentioned wanting to defer this feature)

Yes that would be good, but once we dip our foot into local storage, it becomes difficult to change the structure. That's why I've been hesitant to start with it 😅

packages/vue-client/src/components/GameList.vue
merowin: I agree

benjaminpjones:
> I wasn't sure how that works when user input changes. Do I need to trigger a router navigation? (If so, I think that's a downside imo)

It is a router nav, but it doesn't remount the component. You'd call `router.replace({ query: { ...route.query, page: 2 } })` (or `push`, depending how you want history to work), and read state back from `route.query` (probably through a `computed`).

> But tbh I also don't see much advantage. What would be an example scenario where it makes sense to link to a certain page?

It's a fair point - I don't think it would much be used this way. I am thinking more of the "back button" use-case.  Your solution works for navigation within the app.  Incorporating URL params would make it work for navigation away from govariants.com too.

Also, comparing to Pinia, the advantage is no global state singleton.

> To be honest I didn't know about [KeepAlive].

tbh me too haha.  Just did a search to see what alternatives exist.

> Would it re-fetch the games in the scenario [user creates game, then navigates back to list]?

By default no, because the component isn't remounted - you'd have to explicitly hook `onActivated(() => load())` to re-fetch on return.

> Yes that would be good, but once we dip our foot into local storage, it becomes difficult to change the structure. That's why I've been hesitant to start with it 😅

agreed - we can focus on non-persistent state for now!

PR #501, opened 2026-04-11
Restrict phantom exportState so spectators see an empty board

### LINKED ISSUE #500: Phantom exportState leaks full board to spectators
## Bug

[packages/shared/src/variants/phantom.ts:5-19](../blob/main/packages/shared/src/variants/phantom.ts#L5-L19):

```ts
exportState(player?: number): BadukState {
  const state = super.exportState();
  if (player === undefined) {
    return state;
  }
  // ...filter board to only show own stones
}
```

When `exportState` is called with `player === undefined` (the spectator/broadcast path at [packages/server/src/games.ts:271](../blob/main/packages/server/src/games.ts#L271)), phantom returns the full unfiltered board. A seated player can open a second browser tab as a spectator and see every opponent stone — defeating the fog-of-war.

## Contrast

[packages/shared/src/variants/lighthouse/lighthouse.ts:50-51](../blob/main/packages/shared/src/variants/lighthouse/lighthouse.ts#L50-L51) deliberately restricts the spectator view to the intersection of both players' visibility for exactly this reason.

## Related

Same shape of bug as #496 (parallel staged moves leak to spectators). The underlying issue is that the contract for "what does `exportState(undefined)` return for a hidden-info variant" is undefined across the codebase — might be worth addressing more systematically (e.g. an explicit `hidden_information` flag on the `Variant` interface and a shared spectator-view policy).

benjaminpjones:
## Summary

- Fix spectator leak: `Phantom.exportState(undefined)` previously returned the full unfiltered board, letting a seated player open a second tab as a spectator to reveal opponent stones. Removing the early return makes the existing filter apply to spectators too — `color_to_player` returns `undefined` only for `Color.EMPTY`, so matching against an `undefined` player keeps empty cells and blanks out every stone.
- Add post-game reveal: once `phase === "gameover"`, return the unfiltered board to everyone (seated players and spectators). Phantom is meant to be a reveal at the end — previously seated players continued to see only their own stones forever, making post-game review impossible.
- Tests updated and extended: the old assertions on `exportState()` (no arg) were checking the buggy full-board path; added a new test that plays stones, passes twice, and verifies every caller sees the full board after gameover.

Fixes #500. Tackles the phantom half of #502; the broader "extend `exportState` with a viewer context" design question remains open there.

## Test plan

- [x] `yarn workspace @ogfcommunity/variants-shared test phantom` (3 tests pass)
- [x] Full shared suite passes (192+ tests)
- [x] `yarn lint` clean
- [x] Manual: open a phantom game, confirm spectators see an empty board during play, and full board after both players pass

Screenshot 2026-04-11 at 11 59 13 AM
Screenshot 2026-04-11 at 11 59 22 AM
Screenshot 2026-04-11 at 11 59 32 AM

### DIFF
--- a/packages/shared/src/variants/__tests__/phantom.test.ts
+++ b/packages/shared/src/variants/__tests__/phantom.test.ts
@@ -56,3 +56,24 @@ test("Play a game with captures", () => {
   expect(game.exportState(0).captures).toEqual({ "0": 0, "1": 2 });
   expect(game.exportState(1).captures).toEqual({ "0": 0, "1": 2 });
 });
+
+test("Reveals full board to everyone after the game ends", () => {
+  const game = new Phantom({ width: 4, height: 2, komi: 0.5 });
+  // - W B -
+  // - W B -
+  game.playMove(0, "ca");
+  game.playMove(1, "ba");
+  game.playMove(0, "cb");
+  game.playMove(1, "bb");
+  game.playMove(0, "pass");
+  game.playMove(1, "pass");
+
+  const fullBoard = [
+    [Color.EMPTY, Color.WHITE, Color.BLACK, Color.EMPTY],
+    [Color.EMPTY, Color.WHITE, Color.BLACK, Color.EMPTY],
+  ];
+
+  expect(game.exportState().board).toEqual(fullBoard);
+  expect(game.exportState(0).board).toEqual(fullBoard);
+  expect(game.exportState(1).board).toEqual(fullBoard);
+});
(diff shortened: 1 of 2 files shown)


benjaminpjones:
Added the "gameover reveal" (what lighthouse does)

merowin (approved):
This is nice, thanks!

This whole debate about `exportState` makes me think that we should refine the documentation in `abstract_game.ts` by specifying that the function is supposed to return the observer view when `player === undefined`. Doesn't need to be in this PR though.

// Edit: Just saw that you also had this comment in https://github.com/govariantsteam/govariants/pull/497 so I'll include the docs update there!

PR #506, opened 2026-04-14
read + write list settings in url params

merowin:
#382 

using URL params so settings of the games list are preserved when navigating back in browser history.

### DIFF
--- a/packages/vue-client/src/components/GamesFilterForm.vue
+++ b/packages/vue-client/src/components/GamesFilterForm.vue
@@ -6,19 +6,16 @@ import {
 } from "@ogfcommunity/variants-shared";
 import { ref } from "vue";
 
-const props = defineProps<{ filter: GamesFilter }>();
-
-const emit = defineEmits<{ (e: "update:filter", filter: GamesFilter): void }>();
-function emitFilter() {
-  emit("update:filter", createFilter());
-}
+const filter = defineModel<GamesFilter>();
 
 const variants: string[] = getVariantList();
 
-const variant = ref(props.filter.variant ?? "");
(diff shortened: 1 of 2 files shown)


packages/vue-client/src/components/GamesFilterForm.vue
```diff
@@ -4,31 +4,38 @@ import {
   getVariantList,
   type GamesFilter,
 } from "@ogfcommunity/variants-shared";
-import { type Ref, ref, computed } from "vue";
+import { ref } from "vue";
 
-const emit = defineEmits<{ (e: "filterChange", filter: GamesFilter): void }>();
+const props = defineProps<{ filter: GamesFilter }>();
+
+const emit = defineEmits<{ (e: "update:filter", filter: GamesFilter): void }>();
```
benjaminpjones: for my own understanding, why does the event's name change? is this because of how `v-model` works?

benjaminpjones (approved):
thanks for doing this! The reactivity looks clean :D

packages/vue-client/src/components/GamesFilterForm.vue
merowin: Yes I believe it's how v-model works: https://vueframework.com/guide/migration/v-model.html#_3-x-syntax

Now I also found this: https://vuejs.org/guide/components/v-model.html looks even cleaner

merowin: fyi I changed it to use `defineModel`

benjaminpjones: Cool! And thanks for clarifying

Issue #502, opened 2026-04-11
Reveal full board to players and spectators after hidden-info game ends


benjaminpjones:
## Problem

In [phantom.ts](../blob/main/packages/shared/src/variants/phantom.ts), `exportState` always filters the board — even after the game is over. So:

- Seated players continue to see only their own stones forever after the game ends.
- Spectators continue to see an empty board (after #500 is merged).

In classical phantom go, the end of the game is the reveal - that's half the fun. Full information is critical to post-game review.

## Directions considered

This is the second hidden-info variant to need a gameover-aware view, which suggests the pattern is worth formalizing. Options:

1. **Per-variant `isGameOver` check** — smallest change, each variant handles it internally. Works today but every new hidden-info variant has to remember.
2. **Pass a context object to `exportState`** instead of a bare `player?: number` — e.g. `exportState({ player?: number; phase: Phase })`. Variants opt in to fields they care about. Cleaner, but touches the `AbstractGame` contract and every variant.
3. **Add a separate `exportFinalState()` method** called once on game over. Avoids overloading `exportState` but introduces a second path the server has to know about.

Probably (1) is a reasonable stopgap, with (2) as the long-term fix. Related to the general question raised in #500: there's no shared contract for what `exportState(undefined)` should return for hidden-info variants.

## Scope

- Formalize the pattern in the `Variant` / `AbstractGame` interface.
- Update hidden info variants (parallel variants may not need the fix, since each round displays full info anyway.  partial rounds are an implementation detail)

Update (4/11): edited to emphasize option 2, remove some extraneous content

merowin:
Probably related: #406 

Note that lighthouse reveals the true board state only for the last round. To see what exactly happened throughout the entire game history, a "omniscient" perspective for all past rounds would be neat (but should only be available after game is over)

benjaminpjones:
Oh oops, yeah thats dupe

benjaminpjones:
Regarding the isGameOver - I should have removed that piece. Its a good step forward, but this issue is really about something like (2)

PR #507, opened 2026-04-16
expose hidden info after game is over (#405)

### LINKED ISSUE #406: Phantom Go should show full information after the game has ended
It's impossible to really tell what happened in the game as-is.

benjaminpjones:
## Summary

- Replaces `AbstractGame.exportState(player?: number)` with `exportState(context?: ExportContext)` where `ExportContext = { player?: number; phase: GamePhase }`, so hidden-info variants can reveal the full state during post-game history review.
- The server's `getGameState` and the demo view's `getGame` now replay the full move list first to capture the game's final phase, then rewind for history view, and pass that final phase to `exportState`.
- Phantom and lighthouse switch from `this.phase === "gameover"` to `context.phase === "gameover"` — the internal phase is always `"play"` at a mid-game historical round, which is why review was broken.

Fixes #406. Implements option (2) from #502.

## Test plan

- [x] `yarn build` — clean
- [x] `yarn test` — 235/235 passing
- [x] `yarn lint` — clean
- [x] New phantom test covers the history-review scenario: mid-game position with explicit `phase: "gameover"` reveals the full board
- [x] Manually spot-checked phantom and lighthouse in the browser

### DIFF
--- a/packages/shared/src/abstract_game.ts
+++ b/packages/shared/src/abstract_game.ts
@@ -45,7 +45,7 @@ export abstract class AbstractGame<
    * When omitted, hidden-info variants default to an observer-during-play view
    * (no leakage).
    */
-  abstract exportState(context?: ExportContext): GameState;
+  abstract exportState(context: ExportContext): GameState;
 
   /**
    * Returns the list of players that need to play a move the next round.
@@ -110,5 +110,5 @@ export type GamePhase = "play" | "gameover";
  */
 export type ExportContext = {
   player?: number;
-  phase?: GamePhase;
+  phase: GamePhase;
 };
--- a/packages/shared/src/abstract_game.ts
+++ b/packages/shared/src/abstract_game.ts
@@ -36,9 +36,16 @@ export abstract class AbstractGame<
 
   /** Returns the game state from the perspective of a specific player or observer.
    *
-   * @param player the 0-indexed seat number of a player, or undefined for observers.
+   * @param context.player the 0-indexed seat number of a player, or undefined for observers.
+   * @param context.phase the phase of the overall game. Hidden-info variants can
+   *   use this to reveal the full state when `"gameover"` — including during
+   *   history review of a completed game, where the replayed game object's own
+   *   phase is still `"play"` at the viewed round.
+   *
+   * When omitted, hidden-info variants default to an observer-during-play view
+   * (no leakage).
    */
-  abstract exportState(player?: number): GameState;
+  abstract exportState(context?: ExportContext): GameState;
 
   /**
    * Returns the list of players that need to play a move the next round.
@@ -91,3 +98,16 @@ export abstract class AbstractGame<
 }
 
 export type GamePhase = "play" | "gameover";
+
+/**
+ * Context passed to `exportState` so variants can tailor the view.
+ *
+ * `phase` is the phase of the overall game. When a completed game is being
+ * reviewed at a historical round, the replayed game object's `this.phase` is
+ * still `"play"` at that round — pass `phase: "gameover"` here to let
+ * hidden-info variants reveal the full state.
+ */
+export type ExportContext = {
+  player?: number;
+  phase: GamePhase;
+};
--- a/packages/shared/src/abstract_game.ts
+++ b/packages/shared/src/abstract_game.ts
@@ -112,3 +112,24 @@ export type ExportContext = {
   player?: number;
   phase?: GamePhase;
 };
+
+/**
+ * Shared rule for hidden-info variants deciding whether to reveal the full
+ * state:
+ * - an observer viewing a completed game always sees everything
+ * - a seated player sees everything only when viewing the final state (the
+ *   replayed game object has itself reached `"gameover"`)
+ * - otherwise the caller should return the per-player filtered view
+ *
+ * @param currentPhase the phase of the replayed game object (`this.phase`),
+ *   which differs from `context.phase` during history review
+ */
+export function shouldRevealHiddenInfo(
+  currentPhase: GamePhase,
+  context: ExportContext | undefined,
+): boolean {
+  const gameIsOver = context?.phase === "gameover";
+  const atFinalState = currentPhase === "gameover";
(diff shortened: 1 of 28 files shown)


packages/vue-client/src/views/VariantDemoView.vue
```diff
@@ -76,29 +76,34 @@ function getGame(
...
-    game_obj.playMove(player, move);
+  // Replay all moves first so we know the game's final phase — hidden-info
+  // variants need this to reveal the full state during post-game history
+  // review.
+  const fullGame = makeGameObject(variant, cloneConfig(config));
+  for (const m of moves) {
+    const { player, move } = getOnlyMove(m);
+    fullGame.playMove(player, move);
   }
+  const finalPhase = fullGame.phase;
```
benjaminpjones, line 87: arguably replaying the game just to get the phase is overkill - we can store game phase in the database.  But for now, I didn't want to touch the schema.  Also this approach is less likely to fall out-of-sync.

packages/shared/src/variants/lighthouse/lighthouse.ts
```diff
@@ -46,10 +46,11 @@ export class Lighthouse extends AbstractGame<
     this.board = new LighthouseGrid(config.board.width, config.board.height);
   }
 
-  override exportState(player?: number): LighthouseState {
+  override exportState(context?: ExportContext): LighthouseState {
```
merowin, line 50: Question: Will this make it so players can't see their perspective retrospectively? What if I want to replay my game exactly as I saw it?

benjaminpjones: Your read is correct, and good point 😀

What if we do this instead?

```
player is defined => hide info (except the final state?)
player is undefined => hide info if ongoing game, full info if game has completed
```

merowin: Sounds good to me 🙂

merowin: I've also been thinking it would be neat if, after the game has ended, spectators could replay the game from the players perspective.

benjaminpjones: updated! and filed #508 for your second comment

packages/vue-client/src/views/VariantDemoView.vue
merowin: I agree that storing the game result in the data base can be helpful for optimisation (although the variant demo view wouldn't benefit from it).

While there are ways to optimise the variant demo view performance, it might come at the cost of readability.

packages/shared/src/lib/hidden_info.ts
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * - an observer viewing a completed game always sees everything
+ * - a seated player sees everything only when viewing the final state — the
+ *   replayed game object has itself reached `"gameover"`
+ * - otherwise the caller should return the per-player filtered view
+ *
+ * @param currentPhase the phase of the replayed game object (`this.phase`),
+ *   which differs from `context.phase` during history review
+ */
+export function shouldRevealHiddenInfo(
```
merowin, line 15: This is clean and well documented, nice.

packages/shared/src/variants/rengo.ts
```diff
@@ -59,11 +59,16 @@ export class Rengo<TSubstate extends object> extends AbstractGame<
   numPlayers(): number {
     return sum(this.config.teamSizes);
   }
-  exportState(player?: number): TSubstate {
+  exportState(context?: ExportContext): TSubstate {
+    const phase = context?.phase ?? "play";
```
merowin: Why default `phase` to "play" here? I feel like the principled way would be to pass the input to the subgame unchanged. (it's different for `player` since that needs to be mapped to the respective team)

packages/shared/src/abstract_game.ts
```diff
@@ -36,9 +36,16 @@ export abstract class AbstractGame<
...
+   * @param context.phase the phase of the overall game. Hidden-info variants can
+   *   use this to reveal the full state when `"gameover"` — including during
+   *   history review of a completed game, where the replayed game object's own
+   *   phase is still `"play"` at the viewed round.
+   *
+   * When omitted, hidden-info variants default to an observer-during-play view
+   * (no leakage).
    */
-  abstract exportState(player?: number): GameState;
+  abstract exportState(context?: ExportContext): GameState;
```
merowin: Is there an argument against making the context object required rather than optional? I feel that would be simpler. Same question for `context.phase`

benjaminpjones: so i actually explored this! The downside of required context is that we need to add the parameter to any superclasses of hidden variants (i.e. `Baduk` -> `Phantom`).

Not necessarily a bad thing - let me know what you think is best.

non-null `context.phase` should be doable i think

benjaminpjones: you know what, I tried it, and the blast radius was not that large (just rengo affected) updated to be non-null

Edit: okay now I see what I was seeing earlier.  *If* phantom/one-color is non-null, `Baduk` must be non-null.  Then we have to update any Baduk subvariants which call `super.exportState()`.  Maybe this is fine, but for now I left phantom and one-color nullable.  Let me know if you'd prefer `Baduk` accept non-null context.

Edit 2: I decided to go all the way for consistency (compared to what's already here, it's not a super large blast radius)

merowin: Thanks for exploring this!

Ok I understand the downside now (we see a lot of `exportState({ phase: "play" })`

To clarify why I thought this would be simpler, is because there are less possibilities / code paths to worry about. So it might still be worth it.

If an implementer doesn't know what phase to set as input, it is always ok to enter `phase: gameObject.phase`.

benjaminpjones: Agree - I think the benefits are tangible! (even though we have to update more callsites)

merowin (approved):
This will make it much easier to find out what exactly happened during the game 😁

Thank you!

PR #510, opened 2026-04-23
add vue i18n

merowin:
Adds package [vue i18n](https://www.npmjs.com/package/vue-i18n) and configures it s.t. the browser locale is used.
Basic example usage with interpolation.

### DIFF
--- a/packages/vue-client/package.json
+++ b/packages/vue-client/package.json
@@ -40,6 +40,7 @@
     "voronoijs": "^1.0.0",
     "vue": "^3.4.25",
     "vue-error-boundary": "^2.0.3",
+    "vue-i18n": "^11.3.2",
     "vue-router": "^4.2.2",
     "vue3-chessboard": "^1.2.0"
   },
--- a/packages/vue-client/src/App.vue
+++ b/packages/vue-client/src/App.vue
@@ -31,32 +31,26 @@ const toggleMenuFn = (event: MouseEvent) => {
 
 <template>
   <nav>
-    <RouterLink class="navLogo" to="/"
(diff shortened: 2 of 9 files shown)


benjaminpjones:
Yooo this is glorious!

I know it's draft, but couple questions:

- Is there any way to set/switch between locales? I guess that's at the browser level?
- if I want to add a new language, whats the best way to generate the file? Does vue have any tools to auto create/update these JSON from the source? (Not a requirement, but good to document the "blessed" workflows for potential contributors)

merowin:
> * Is there any way to set/switch between locales? I guess that's at the browser level?

Yes this is possible, see e.g. https://vue-i18n.intlify.dev/guide/essentials/scope.html

At the browser level the user can switch locale, but I believe it may require a reload.

> * if I want to add a new language, whats the best way to generate the file? Does vue have any tools to auto create/update these JSON from the source? (Not a requirement, but good to document the "blessed" workflows for potential contributors)

I'm not aware of a tool that can generate such a file. I think a good way to add a new language is: copy `en.json` and translate keys incrementally.

But there are IDE plugins that can help work with such translation files. They can help e.g. adding a new key / update a key in all files.

merowin:
About the tooling, I tried [i18n-ally](https://github.com/lokalise/i18n-ally)

it allows you to add and edit all translations within the html template, which is very nice.

There were some strange things though. In addition to the normal setup steps, I had to add Vue as framework in the i18n-ally settings to make it work. When I edited a key, it stumpled trying to apply the change where the $t is called.

I think a tool like this is very helpful once there are many languages / translation keys to keep track of. For now I'd prefer to manually edit the jsons though.

PR #511, opened 2026-04-30
Add Rizoma variant: multi-board Go with capture transfer

VRM-Valerio:
Rizoma is a two-or-more-board Go variant where capturing stones gives the capturing player a choice: keep them as prisoners for scoring, or redeploy them as bonus moves on any other board.

Game logic (packages/shared):
- New variant: rizoma.ts with full Go rules (captures, ko, suicide check) on each board independently, capture-transfer mechanic, dynamic komi (second opener per board), global komi for White, territory scoring summed across boards, per-board score breakdown in exported state
- Config: per-board { width, height, komi } + global_komi field
- movePreview support for submit-move UX
- sanitizeConfig for forward compatibility with old configs
- rulesDescription with full variant rules
- Test suite: rizoma.test.ts (7 tests)

UI (packages/vue-client):
- RizomaBoard.vue: side-by-side board layout, transfer badges, komi and score labels per board, source-board click-blocking during transfers
- RizomaConfigForm.vue: per-board config form (up to 5 boards, each with width/height/komi), global komi field
- Registered in board_map, config_form_map, variant_map, index exports

Server (packages/server):
- games.ts: add turn-order validation before new moves in handleMoveAndTime (not during move replay) to enforce B/W alternation server-side

Fixes (packages/vue-client):
- GameView.vue: guard RouterLink with v-if="variant" to prevent "Missing required param 'variant'" warning during initial load

### DIFF
--- a/packages/server/src/games.ts
+++ b/packages/server/src/games.ts
@@ -185,15 +185,6 @@ export async function handleMoveAndTime(
 
   const { player: playerNr, move: new_move } = getOnlyMove(moves);
 
-  const nextPlayers = game_obj.nextToPlay();
-  if (
-    nextPlayers.length > 0 &&
-    new_move !== "resign" &&
-    !nextPlayers.includes(playerNr)
-  ) {
-    throw new Error(`It is not player ${playerNr}'s turn`);
-  }
-
   const previousRound = game_obj.round;
   game_obj.playMove(playerNr, new_move);
   const isRoundTransition = previousRound !== game_obj.round;
--- a/packages/server/src/games.ts
+++ b/packages/server/src/games.ts
@@ -185,6 +185,15 @@ export async function handleMoveAndTime(
 
   const { player: playerNr, move: new_move } = getOnlyMove(moves);
 
+  const nextPlayers = game_obj.nextToPlay();
+  if (
+    nextPlayers.length > 0 &&
+    new_move !== "resign" &&
+    !nextPlayers.includes(playerNr)
+  ) {
+    throw new Error(`It is not player ${playerNr}'s turn`);
+  }
+
   const previousRound = game_obj.round;
   game_obj.playMove(playerNr, new_move);
   const isRoundTransition = previousRound !== game_obj.round;
--- a/packages/shared/src/variants/__tests__/rizoma.test.ts
+++ b/packages/shared/src/variants/__tests__/rizoma.test.ts
@@ -103,10 +103,10 @@ test("Keep prisoners does not count toward double-pass game end", () => {
   // Player 0 should have 1 pending transfer.
   expect(game.exportState().pending_transfers).toBe(1);
 
-  // Player 0 keeps prisoners (pass with pending_transfers > 0).
-  game.playMove(0, "pass");
+  // Player 0 keeps prisoners via the dedicated move.
+  game.playMove(0, "keep_prisoners");
 
-  // Game must NOT be over — keeping prisoners is not a game-ending pass.
+  // Game must NOT be over — keep_prisoners is not a game-ending pass.
   expect(game.phase).not.toBe("gameover");
   expect(game.exportState().pending_transfers).toBe(0);
 });
--- a/packages/shared/src/variants/__tests__/rizoma.test.ts
+++ b/packages/shared/src/variants/__tests__/rizoma.test.ts
@@ -1,0 +1,151 @@
+import { Rizoma } from "../rizoma";
+
+const B = 0;
+const _ = null;
+const W = 1;
+
+const twoBoard4x4 = {
+  boards: [
+    { width: 4, height: 4, komi: 0 },
+    { width: 4, height: 4, komi: 0 },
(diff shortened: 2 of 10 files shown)


packages/server/src/games.ts
```diff
@@ -185,6 +185,15 @@ export async function handleMoveAndTime(
 
   const { player: playerNr, move: new_move } = getOnlyMove(moves);
 
+  const nextPlayers = game_obj.nextToPlay();
```
merowin: This change makes sense semantically and should work for sequential variants. But variants with parallel moves currently allow users to change their staged move by submitting a new move. Also users who already have staged a move are not listed in game.nextToPlay. So this check will guard against changing staged moves in parallel variants.

My recommendation: Move the check into Rizoma's playMove method

packages/shared/src/variants/__tests__/rizoma.test.ts
```diff
@@ -0,0 +1,151 @@
...
+
+const twoBoard4x4 = {
+  boards: [
+    { width: 4, height: 4, komi: 0 },
+    { width: 4, height: 4, komi: 0 },
+  ],
+  global_komi: 0,
+};
+
+test("Play a game on board 0", () => {
```
merowin, line 15: Thank you for adding test cases!

packages/shared/src/variants/rizoma.ts
```diff
@@ -0,0 +1,475 @@
...
+    ]) {
+      const nx = coord.x + dx;
+      const ny = coord.y + dy;
+      if (nx >= 0 && nx < width && ny >= 0 && ny < height)
+        result.push(new Coordinate(nx, ny));
+    }
+    return result;
+  }
+
+  private getGroup(
```
merowin: I believe there is a utils function that can be reused here.

packages/shared/src/variants/rizoma.ts
```diff
@@ -0,0 +1,475 @@
...
+    return { black, white };
+  }
+
+  override numPlayers(): number {
+    return 2;
+  }
+
+  override specialMoves(): { [key: string]: string } {
+    return {
+      pass: this.pending_transfers > 0 ? "Keep prisoners" : "Pass",
```
merowin: This doesn't seem right. "pass" shouldn't be the move to keep prisoners.

My recommendation: 

`return this.pending_transfers > 0 ? { "keep_prisoners" : "Keep prisoners", resign: "Resign" } : { pass: "Pass", resign: "Resign" }

Then the move is encoded as "keep_prisoners", so the playMove method needs to be adjusted accordingly. This makes it easier to read and reason about, for people who read the encoded moves.

packages/shared/src/variants/rizoma.ts
```diff
@@ -0,0 +1,475 @@
...
+  defaultConfig(): RizomaConfig {
+    return {
+      boards: [
+        { width: 7, height: 7, komi: 0 },
+        { width: 7, height: 7, komi: 0 },
+      ],
+      global_komi: 0.5,
+    };
+  },
+  sanitizeConfig(config: object): RizomaConfig {
```
merowin: sanitizeConfig is only needed when the config shape changes while there are already game documents in the data base. We can and should remove it here to simplify.

packages/vue-client/src/board_map.ts
```diff
@@ -16,6 +17,7 @@ const board_map: {
   keima: KeimaBoard,
   quantum: QuantumBoard,
   rengo: PolymorphicBoard,
+  rizoma: RizomaBoard as Component<{ config: unknown; gamestate: unknown }>,
```
merowin: The type assertion (`as Component<...>`) should not be necessary here. Is there a reason, or can we remove it?

packages/vue-client/src/views/GameView.vue
```diff
@@ -302,7 +302,7 @@ const createTimeControlPreview = (
         <NavButtons v-model="view_round" :game-round="current_round" />
 
         <div id="variant-info">
-          <div>
+          <div v-if="variant">
```
benjaminpjones: This one seems unrelated to rizoma implementation.  can we remove it from the PR?

benjaminpjones (changes requested):
This variant seems really cool! I tried it locally and seems to work as intended.  Very cool that boards can be different sizes :)

Generally, the rizoma code looks good, but I see there were changes made to the framework

- games.ts, which @merowin commented, is the big one - **this change will break other variants**
- GameView.vue we should probably also roll back just to keep the PR scoped (let me know if it was important, we can add it separately)

VRM-Valerio:
Thank you both for the thorough review and for trying it out locally! I've addressed all the change requests. I did my best to keep things clean, though I was away for a bit. Happy to make any further adjustments if needed — just let me know.

merowin (approved):
Looks good to me, thank you for the contribution!

🙂

benjaminpjones (approved):

PR #515, opened 2026-06-03
Fix board shape/variant config desync in game creation form

benjaminpjones:
## Problem

Selecting a board shape (e.g. a circular board) and then switching the variant to another Baduk-derived variant (Tetris, Capture, Phantom, Freeze, etc.) shows the chosen board in the **preview**, but the created game uses the **default grid** instead. Reported on reddit: https://www.reddit.com/r/baduk/comments/1ttzksz/comment/opccy6a/

## Root cause

When switching between variants, if the selected variant uses the same config form, `<component :is="variantConfigForm">` **reuses the instance** across variant changes. Its `setup` never re-runs and it never re-emitted. Result: the displayed/previewed board diverged from the config that is actually sent to the server.

## Fix

```vue
<component :is="variantConfigForm" :key="variant" ... />
```

Keying on `variant` remounts the config form on variant change, so it re-emits its config and the preview stays in sync with what gets created. Switching variant now resets the board to that variant's default, matching the underlying data that gets sent to the server.

## Changes

- bug fix (`:key`)
- Test to catch the bug
- add @testing-library/vue for more semantic tests (`getByLabelText`, `getByText`)

## Test plan

1) select "circular"
2) select "tetris"
3) observe that config form resets to the default "grid"
4) start game - observe that it's grid
    - alternatively, select "circular" and it will be circular

### DIFF
--- a/packages/vue-client/src/components/GameCreation/GameCreationForm.vue
+++ b/packages/vue-client/src/components/GameCreation/GameCreationForm.vue
@@ -91,8 +91,8 @@ const hasTimeConfigForm = computed(
     <h3>Create a game</h3>
     <div>
       <div class="col">
-        <label>Variant</label>
-        <select v-model="variant">
+        <label for="variant-select">Variant</label>
+        <select id="variant-select" v-model="variant">
           <option
             v-for="variantOption in variants"
             :key="variantOption"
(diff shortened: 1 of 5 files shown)


packages/vue-client/src/components/GameCreation/GameCreationForm.vue
```diff
@@ -107,6 +107,7 @@ const hasTimeConfigForm = computed(
     <template v-if="variantConfigForm">
       <component
         :is="variantConfigForm"
+        :key="variant"
```
benjaminpjones, line 110: this is the bug fix.  all other changes are in service of testing.

PR #552, opened 2026-08-23
Persist immediate submit preference in localStorage

benjaminpjones:
## Summary

The "immediate submit" checkbox in `GameView` reset on every page load. This stores it in `localStorage` so it sticks.

`localStorage` rather than the user account is deliberate: this is a client-specific preference, and someone's answer on their phone may differ from their answer on desktop.

## Also included: a small pre-existing bug fix

Added a `watch` that drops the preview when "immediately submit" is switched on.

## Testing

Ensure it's sticky

1) go to a game
2) enable **immediate submit**
3) leave page, come back
4) ensure it's still set to **immediate submit**
5) do the same for **immediate submit OFF**

Ensure immediate submit clears any pending move

1) go to game
2) stage a move
3) enable **immediate submit**, observe move is no longer staged

### DIFF
--- a/packages/vue-client/src/views/GameView.vue
+++ b/packages/vue-client/src/views/GameView.vue
@@ -20,6 +20,7 @@ import {
   getMovePreview,
 } from "@govariants/shared";
 import { computed, ref, watch, watchEffect, type Ref } from "vue";
+import { useLocalStorage } from "@vueuse/core";
 import { socket } from "../requests";
 import NavButtons from "@/components/GameView/NavButtons.vue";
 import PlayersToMove from "@/components/GameView/PlayersToMove.vue";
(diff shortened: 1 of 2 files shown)


packages/vue-client/src/views/GameView.vue
```diff
@@ -71,7 +72,13 @@ const adminMode = ref<boolean>(false);
 const creator = ref<User | undefined>();
 const subscription = ref<NotificationType[]>([]);
 
-const userEnabledImmediateSubmit = ref(false);
+// This is a client-specific preference: someone may want immediate submit
+// on desktop but not on their phone, so it lives in localStorage rather
+// than on the user account.
+const userEnabledImmediateSubmit = useLocalStorage(
+  "govariants:immediate-submit",
```
merowin: I think we should pull the key into a file dedicated to local storage keys. Once we add more local storage keys, that'll make it easier to find the available keys and reuse them.

merowin (approved):
Looks good to me (one potential improvement commented above)

thanks!

packages/vue-client/src/views/GameView.vue
benjaminpjones: Agreed and updated! This will also be useful if we ever move this into a dedicated settings page.

We can go even further by centralizing defaults or wrapping `useLocalStorage` to make TypeScript understand the schema.  I deferred these potential improvements for when we have more settings to reason about.

PR #555, opened 2026-08-29
Push notification counts over the socket

benjaminpjones:
## Problem

The bell badge showed the unread count as of page load, but did not update as the count changed.

`NotificationsNav` fetches `/notifications/count` inside a `watchEffect` whose only reactive dependencies are `user.value` and `store.csrf_token`, so it fires once after login and never again. There is no socket event for the count, and the component is permanently mounted in `App.vue` (the hamburger toggles a CSS class, not `v-if`), so no navigation remounts it.

## Change

Broadcast each affected user's unread count whenever their notifications change.

- Sockets join a per-user room, `user/<id>/notifications`, in the `connection` handler.
- Counts are broadcast at the end of each operation that mutates notifications.
- `NotificationsNav` listens for the event; the existing fetch stays as the page-load value.

## Verification

`yarn build`, `yarn test` and `yarn lint` are clean. Beyond that, run against the real stack with two registered users and a live socket:

```
A plays (A is black, nothing pending)  << push: 0
B plays -> A is on the move            << push: 1
A plays -> A has addressed it          << push: 0
```

Added 5 integration tests for the broadcasts (`notifications.test.ts`), a rejection test for user topics, and `to()` on the socket.io mock.

## Follow-ups, deliberately not in this PR

- **Socket identity is fixed at the handshake.** see #557

- **A race on the final move.** see #558 

https://claude.ai/code/session_019UDbeZZ4okUf2BrXLVyFuV

### DIFF
--- a/packages/server/src/notifications/notifications.ts
+++ b/packages/server/src/notifications/notifications.ts
@@ -4,9 +4,12 @@ import { UserNotifications } from "./notifications.types";
 import {
   GameNotification,
   GameSubscriptions,
+  NOTIFICATIONS_COUNT_EVENT,
   Notifications,
   NotificationType,
 } from "@govariants/shared";
+import { io } from "../socket_io";
+import { userNotificationsTopic } from "../socket_validation";
 
 function outwardMap(userNotifications: UserNotifications): GameNotification[] {
   return userNotifications.notifications as GameNotification[];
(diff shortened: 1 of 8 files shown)


packages/server/src/notifications/notifications.ts
```diff
@@ -54,7 +71,24 @@ export async function getUserNotificationsCount(
     ])
     .toArray();
 
-  return queryResult.at(0)?.num ?? 0;
+  for (const { userId, num } of queryResult) {
```
merowin: nit: why convert to map? Seems like returning the array would be sufficient.

merowin (approved):
Looks good to me, thanks! Had one question but nothing thats blocking.

packages/server/src/notifications/notifications.ts
benjaminpjones: I think an array is possible, but this conversion ensures is a row for every user.  As-is, notifications.aggregate() will omit a user's notifications if there is no document. It doesn't matter now because delete all only happens for deleted users, but I think this is safer for the future.

As for the return shape, I think it's slightly more ergonomic to provide the Map keyed on userId rather than have the caller scan an array or infer by index.  But for existing call sites, that's not an issue (one has only one element, the other iterates the whole container).

PR #559, opened 2026-08-31
Keep the notification bell in the top bar on mobile

benjaminpjones:
## Problem

On narrow viewports the nav links collapse behind the hamburger, and `NotificationsNav` lived inside that collapsed menu. The unread count was fetched and stored, but the badge was invisible until the menu was opened — there was no way to tell you had notifications at a glance.

## Change

The bell now renders in the nav bar itself on mobile, beside the hamburger, and there is no Notifications entry in the menu. Desktop is unchanged: the bell stays with the other nav links, after Rules.

| | before | after |
|---|---|---|
| mobile bar | logo … hamburger | logo … 🔔 hamburger |
| mobile menu | Home / About / Rules / **Notifications** / user | Home / About / Rules / user |
| desktop | Home / About / Rules / 🔔 … user | unchanged |

### Why the component is rendered twice

The two placements are genuinely different DOM positions — on mobile the bell has to sit *outside* `.navContent` (the element the hamburger collapses), on desktop *inside* it — and CSS can't move a node between parents. So `<NotificationsNav>` appears in both positions and a single media query picks which copy is visible. It's the same `@media (max-width: 768px)` block that already toggles the hamburger, so the two can't drift apart.

Alternatives considered and rejected:

- **`display: contents` + `order`** on the nav children. Works with one instance, but requires an explicit `order` on every nav child — a trap for whoever adds the next nav item.
- **`<Teleport>`.** One instance and exact placement, but the target lives inside `App.vue` itself and Vue mounts an element's children before inserting it into the document, so the target isn't reachable at root mount. It also needs `useMediaQuery`, i.e. a third copy of the 768px breakpoint — in JS this time, where drifting from the CSS breaks the layout silently.

The duplication is only safe because the `/notifications/count` fetch moved up to `App.vue` first, so it runs once regardless of how many bells are mounted. `NotificationsNav` is now purely presentational.

### Incidental cleanups

The bell is icon-only in both layouts now (in the bar it's a bar icon, not a menu row), so the `.mobile-only` label is gone. That removes two things with it:

- the badge overlapping the label's first letter, which it did on mobile whenever the count was non-zero (`right: -15px` is tuned for desktop's `1.2rem` item padding);
- a second hardcoded `768px` media query carrying a "keep in sync with App.vue" comment.

Since that label was the component's only translated string, the accessible name moved to i18n (`notifications-unread`, added to `en.json` and `de.json`) rather than leaving hardcoded English as the only thing a screen reader gets.

## Testing

`lint:check`, `type-check` and `vitest` all pass.

Layout verified against the dev server at 390px and 1100px, logged out, with the unread count forced to 7 so the badge renders.

### Mobile, menu closed (390px)

This is the bug: seven unread, and before the change nothing in the bar says so.

Before:

After:

### Mobile, menu open (390px)

Note the badge landing on the "N" of Notifications in the before shot, and the absence of the row in the after shot.

Before:

After:

### Desktop (1100px)

Unchanged — one bell, still after Rules.

Before:

After:

## Deliberately not included

- **Bar icon sizing.** The bell renders at 14×16px in both layouts (measured — this PR doesn't scale it), but in the bar its only neighbour is the 40×40px hamburger, so it reads as undersized. The hamburger is sized with `calc(var(--navbar-height) * 0.8)`, the same multiplier as the logo. Worth sizing both bar icons together as a follow-up.
- `right: -15px` on the badge, left as-is: it's tuned for desktop's padding, and in the mobile bar it drifts slightly toward the hamburger. Changing it would move desktop pixels.
- Logged-out users still see a bell. That's pre-existing desktop behavior, now just more prominent.
- The nav's accessibility gaps (the hamburger has no `aria-label` / `aria-expanded`, Escape doesn't close the menu) and its manual `document` click listener. Both are real, both are out of scope here.

### DIFF
--- a/packages/vue-client/src/App.vue
+++ b/packages/vue-client/src/App.vue
@@ -2,15 +2,32 @@
 import { RouterLink, RouterView } from "vue-router";
 import UserNav from "./components/UserNav.vue";
 import NotificationsNav from "./components/NotificationsNav.vue";
-import { ref } from "vue";
+import { ref, watchEffect } from "vue";
 import { library } from "@fortawesome/fontawesome-svg-core";
 import { FontAwesomeIcon } from "@fortawesome/vue-fontawesome";
 import { faBars, faBook } from "@fortawesome/free-solid-svg-icons";
 import { faHouse } from "@fortawesome/free-solid-svg-icons";
 import { faCircleInfo } from "@fortawesome/free-solid-svg-icons";
+import * as requests from "@/requests";
+import { useCurrentUser, useStore } from "@/stores/user";
+import { setNotificationsCount } from "@/stores/notifications";
 
 library.add(faBars, faHouse, faCircleInfo, faBook);
 const is_menu_closed = ref(true);
+
+// The bell is rendered twice (see below), so the unread count is fetched here
+// rather than in NotificationsNav: one request no matter how many are mounted.
+const user_store = useStore();
+const user = useCurrentUser();
+
+watchEffect(async () => {
+  if (user.value && user_store.csrf_token) {
+    await requests
+      .get("/notifications/count")
+      .then((result) => setNotificationsCount(result.count))
+      .catch(alert);
+  }
+});
 
 const closeMenuFn = (event: MouseEvent) => {
   event.stopPropagation();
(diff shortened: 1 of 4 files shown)


benjaminpjones:
Closing this - seems like this PR touched more than layout.  Going to refactor the data fetching a bit first.

PR #560, opened 2026-09-02
[codex] add Border War variant (战争号角-边境线)

longgege-cd:
## What

Add a new variant **Border War（战争号角-边境线）** to govariants.

- New `BorderWar` game class at `packages/shared/src/variants/border_war.ts`
- Registered as `"border war"` in `variant_map.ts`
- Vitest coverage at `packages/shared/src/variants/__tests__/border_war.test.ts`

## Why

Players fight across a central borderline (row 10). Placing stones, capturing, and
scoring all follow the Border War rulebook v9.0 — a Go-like territory variant where
the goal is not to surround the most territory, but to score by **attacking** into the
opponent's half of the board.

## Rules implemented

- **Deploy phase (first 4 moves)** — each side must place their 2 stones on their own
  territory; those stones become **strongholds**.
- **Attack vs. defense zones** — territory, siege, and capture only score in the
  opponent's territory or on the borderline (the "attack zone"). Playing on your own
  side scores nothing but can **replenish troops**.
- **Strongholds (据点)** — capturing an enemy stronghold grants **+10**, does **not**
  count as war damage, and does not give capture score.
- **Scoring** (per the rulebook): capture +4/stone (attack zone), siege +3/stone
  (attack zone), false-eye/territory +2/point where applicable, casualty −1/stone,
  white komi +5 at game end.
- **Piece limit 90 / side**, **komi 5**.
- **Game end** — double pass (both sides resign in turn) triggers final scoring.

## Validation

- `tsc --noEmit` clean
- `vitest run`: all Border War tests pass; full shared suite (38 files / 226 cases)
  passes with no regression
- `prettier --check` / `eslint` clean on the touched files

## Notes for maintainers

- This is a self-contained implementation reusing govariants' `Grid`/`Coordinate`/SGF
  infrastructure; it does not extend `Baduk`, keeping the Border War state machine
  explicit.

### DIFF
--- a/packages/shared/src/variants/__tests__/border_war.test.ts
+++ b/packages/shared/src/variants/__tests__/border_war.test.ts
@@ -128,4 +128,24 @@ describe("战争号角-边境线", () => {
     expect(g.result).toMatch(/^(B\+|W\+|Tie)/);
     expect(typeof g.numeric_result).toBe("number");
   });
+
+  test("劫规则：立即回提还原局面被禁止（superko）", () => {
+    const g = newGame();
+    // 布局（据点远离角部劫争区）：黑 if/jf，白 il/jl
+    g.playMove(0, "if");
+    g.playMove(1, "il");
+    g.playMove(0, "jf");
+    g.playMove(1, "jl");
+    // 搭一个角部单点劫：黑 bb(1,1)、ca(2,0)，白 ab(0,1)、ba(1,0)
+    g.playMove(0, "bb");
+    g.playMove(1, "ab");
+    g.playMove(0, "ca");
+    g.playMove(1, "ba"); // 白 ba 只剩 (0,0) 一气
+    g.playMove(0, "aa"); // 黑 aa 提掉白 ba
+    // 白立即回提 ba 会还原落子 aa 前的局面 → 应抛错
+    expect(() => g.playMove(1, "ba")).toThrow(/repeated/);
+    // 白改下别处（如边境）仍可继续
+    g.playMove(1, "aj");
+    expect(g.phase).not.toBe("gameover");
+  });
 });
--- a/packages/shared/src/variants/__tests__/border_war.test.ts
+++ b/packages/shared/src/variants/__tests__/border_war.test.ts
@@ -70,7 +70,7 @@ describe("战争号角-边境线", () => {
     expect(st.strongholds[1]).toHaveLength(0);
     expect(st.score.stronghold[0]).toBe(20); // 2 × +10
     expect(st.score.casualty[1]).toBe(0); // 不计战损
-    expect(st.score.territory[0]).toBe(0); // 据点提走不计吃子分
+    expect(st.score.captures[0]).toBe(0); // 据点提走不计吃子分
     // 双虚终局
     g.playMove(1, "pass");
     g.playMove(0, "pass");
@@ -92,7 +92,7 @@ describe("战争号角-边境线", () => {
     g.playMove(1, "pass");
     g.playMove(0, "bi"); // 提掉白 bj
     const st = g.exportState({ phase: "play" });
-    expect(st.score.territory[0]).toBe(4); // +4 吃子分
+    expect(st.score.captures[0]).toBe(4); // +4 吃子分
     expect(st.score.casualty[1]).toBe(-1); // 战损
     // 双虚终局
     g.playMove(1, "pass");
@@ -105,5 +105,27 @@ describe("战争号角-边境线", () => {
     expect(g.exportState({ phase: "play" }).pieces_left[0]).toBe(90);
     g.playMove(0, "aa");
     expect(g.exportState({ phase: "play" }).pieces_left[0]).toBe(89);
+  });
+
+  test("双虚终局：finalizeScore 产出有效 g.result", () => {
+    const g = newGame();
+    g.playMove(0, "aa");
+    g.playMove(1, "ak");
+    g.playMove(0, "ab");
+    g.playMove(1, "al");
+    g.playMove(0, "aj");
+    g.playMove(1, "pass");
+    g.playMove(0, "bk");
+    g.playMove(1, "pass");
+    g.playMove(0, "bl");
+    g.playMove(1, "pass");
+    g.playMove(0, "am"); // 提掉白据点 {ak,al}
+    g.result = undefined; // 清零，确认确实走 finalizeScore 而非残留
+    g.playMove(1, "pass");
+    g.playMove(0, "pass"); // 双虚 → finalizeScore
+    expect(g.phase).toBe("gameover");
+    expect(typeof g.result).toBe("string");
+    expect(g.result).toMatch(/^(B\+|W\+|Tie)/);
+    expect(typeof g.numeric_result).toBe("number");
   });
 });
--- a/packages/shared/src/variants/__tests__/border_war.test.ts
+++ b/packages/shared/src/variants/__tests__/border_war.test.ts
@@ -32,6 +32,25 @@ describe("战争号角-边境线", () => {
     expect(() => g.playMove(0, "aj")).toThrow(/own territory/);
   });
 
+  test("防御区提吃普通子：回补兵力 +1", () => {
+    const g = newGame();
+    // 布局：黑 aa、ab，白 ak、al（都是据点）
+    g.playMove(0, "aa");
+    g.playMove(1, "ak");
+    g.playMove(0, "ab");
+    g.playMove(1, "al");
+    // 黑在自己领土 bf(1,5) 布点，白落 a5(0,5)；黑再填 ae(0,4)、ag(0,6) 提掉白 a5
+    g.playMove(0, "bf");
+    g.playMove(1, "af");
+    g.playMove(0, "ae");
+    g.playMove(1, "pass");
+    g.playMove(0, "ag"); // a5 四气（ae/ag/bf 已是黑）→ 提走白 a5，黑防御区补兵 +1
+    const st = g.exportState({ phase: "play" });
+    // 黑占用 = aa+ab+bf+ae+ag = 5，补兵 +1 → 剩余兵力 = 90 - 5 + 1 = 86
+    expect(st.pieces_left[0]).toBe(86);
+    expect(st.score.casualty[1]).toBe(-1); // 白普通子战损
+  });
+
   test("提吃据点：+10 奖励，不计战损、不计吃子分", () => {
     const g = newGame();
     // 布局：白据点 ak(x0,y10)、al(x0,y11)
--- a/packages/shared/src/variants/__tests__/border_war.test.ts
+++ b/packages/shared/src/variants/__tests__/border_war.test.ts
@@ -1,0 +1,90 @@
+import { describe, expect, test } from "vitest";
+import { BorderWar } from "../border_war";
+
+// 19×19 默认棋盘。边境线 BORDER_ROW=9（0基第10行）：
+//   黑领土 row 0..8（sgf 行字符 a..j 中的 a..i 恰好也对应 x 字母，需用坐标 x0y 形式）
+//   边境    row 9  （字符 j）
+//   白领土  row 10..18（字符 k..s）
+const newGame = () => new BorderWar({ ...BorderWar.defaultConfig() });
+
+describe("战争号角-边境线", () => {
+  test("布局阶段前4手自动成为据点", () => {
+    const g = newGame();
+    g.playMove(0, "aa"); // 黑 row0
+    g.playMove(1, "ak"); // 白 row10
+    g.playMove(0, "ab"); // 黑 row1
+    g.playMove(1, "al"); // 白 row11
+    const st = g.exportState({ phase: "play" });
+    expect(st.deploy_done).toBe(true);
+    expect(st.strongholds[0].sort()).toEqual(["aa", "ab"]);
+    expect(st.strongholds[1].sort()).toEqual(["ak", "al"]);
+  });
+
+  test("布局阶段不能在对方领土落子", () => {
+    const g = newGame();
+    // 黑第1手落到白领土（row10）应抛错
+    expect(() => g.playMove(0, "ak")).toThrow(/own territory/);
+  });
+
+  test("布局阶段不能在边境线落子", () => {
+    const g = newGame();
+    // 黑第1手落到边境（row9）
+    expect(() => g.playMove(0, "aj")).toThrow(/own territory/);
+  });
+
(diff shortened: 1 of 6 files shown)


packages/shared/src/variants/border_war.ts
```diff
@@ -0,0 +1,514 @@
...
+  group: Array<{ x: number; y: number }>,
+  board: Grid<Color>,
+): boolean {
+  return getOuterBorder(group, board).some((p) => board.at(p) === Color.EMPTY);
+}
+
+export const borderWarVariant: Variant<BorderWarConfig, BorderWarState> = {
+  gameClass: BorderWar,
+  description:
+    "战争号角-边境线 v9.0\n 双方在风云边境线布阵，前4手建立据点；在对方领土/边境围空、围困、吃子得分，己方领土围困/吃子可补兵；破坏对方已得分包围圈得防御分。",
```
benjaminpjones: I think we should probably have user-facing text be English.  We do offer some support for translation, and Chinese would be a welcome addition, but at this point in time, only English and German are supported.

cc: @merowin, who added i18n support

benjaminpjones (commented):
Cool, thank you for this! I can take a closer look tonight, but it's likely okay given that it just adds a new variant.

Just wanted to flag the translation thing for the user-facing text early. Code comments are probably okay as-is, unless you intend for the rest of the team to read them.

packages/shared/src/variants/border_war.ts
longgege-cd: Thanks for the feedback — I've updated the user-facing text (description and the rule title) to English in commit 284a3b7. Chinese remains only in code comments; the variant's player-facing strings are now English. Happy to add a German translation too if useful later.

longgege-cd:
Thanks! Looking forward to your review. And good to know — I'll keep the code comments as-is.

packages/shared/src/variants/border_war.ts
```diff
@@ -0,0 +1,514 @@
...
+      false,
+    );
+    this.board.forEach((color, pos) => {
+      if (color !== Color.EMPTY || visited.at(pos)) return;
+      const group = getGroup(pos, this.board);
+      group.forEach((p) => visited.set(p, true));
+      const border = getOuterBorder(group, this.board);
+      const borderColors = border.map((p) => this.board.at(p) as Color);
+      const owner = borderColors[0];
+      if (borderColors.every((c) => c === owner)) {
```
benjaminpjones, line 378: nit/edge case: `owner` may be `undefined` on an empty board.  the next line assigns ownership to `WHITE`, meaning that two passes on an empty board assigns white the whole board.

Image

packages/shared/src/variants/border_war.ts
```diff
@@ -0,0 +1,514 @@
...
+    bk_total += enc[0];
+    wt_total += enc[1];
+
+    // 围困分：被围困（无气）棋子仍存盘 → 围困方得分 +3/子（攻击区）
+    // 在 border 中围困分数计入占领分，由 finalize 统一累加
+    const siege = this.computeSiege();
+    bk_total += siege[0];
+    wt_total += siege[1];
+
+    bk_total += this.config.komi === this.config.komi ? 0 : 0; // 贴目记白
```
benjaminpjones: I don't think this line does anything

packages/shared/src/variants/border_war.ts
```diff
@@ -0,0 +1,514 @@
...
+
+export enum Color {
+  EMPTY = 0,
+  BLACK = 1,
+  WHITE = 2,
+}
+
+// ---- 常量（对应引擎 Const.ts）----
+export const BOARD_SIZE = 19;
+export const BORDER_ROW = 9; // 0基：第10行 = 边境线
```
benjaminpjones, line 29: [please fix]

It looks like it's still possible to create non-19 sized boards.

If these are hard-coded, it would be good to assert these in the constructor.  For the GoVariants framework, throwing from the constructor represents an invalid config and the game will not be created.

Two options here:

Error out of the constructor (easiest):

```
const { width, height } = this.config.board;
if (width !== BOARD_SIZE || height !== BOARD_SIZE) {
  throw new Error(
    `BorderWar only supports ${BOARD_SIZE}x${BOARD_SIZE} boards (got ${width}x${height})`,
  );
}
```

Support different sizes (best user experience):

```
if (width > 52 || height > 52) {
  throw new Error("BorderWar does not support sizes greater than 52");
}
if (height < 5 || height % 2 === 0) {
  throw new Error(
    `BorderWar needs an odd height of at least 5 so both sides get territory (got ${height})`,
  );
}
// this member replaces the hard-coded BORDER_ROW
this.border_row = (height - 1) / 2;
```

packages/shared/src/variants/border_war.ts
```diff
@@ -0,0 +1,514 @@
...
+    if (move !== "pass") {
+      const pos = Coordinate.fromSgfRepr(move);
+      if (!this.board.isInBounds(pos)) {
+        throw Error(`Move out of bounds. (move: ${move})`);
+      }
+      if (this.board.at(pos) !== Color.EMPTY) {
+        throw Error(
+          `Cannot place a stone on top of an existing stone. (${move})`,
+        );
+      }
```
benjaminpjones, line 173: [question] is there a ko rule for this variant?

Just FYI there's a [`KoDetector`](https://github.com/govariantsteam/govariants/blob/main/packages/shared/src/lib/ko_detector.ts) utility if you want it.  Your call, just wanted you to be aware.

benjaminpjones (approved):
Sorry for the delay on this review.  I think this looks good to merge, but I did find some things which are worth your review before I do.  Let me know what you think - only the config validation is blocking in my opinion.

tested on my computer - this looks like a cool variant!

benjaminpjones:
Also for game creation - this variant does not have a ConfigForm.  That's okay, but makes a difference in how this shows up to users:

Without ConfigForm (current) shows raw JSON:

Screenshot 2026-09-04 at 5 30 18 AM

With ConfigForm, it's a GUI which can guide your users to valid configs:

Screenshot 2026-09-04 at 5 29 36 AM

packages/shared/src/variants/border_war.ts
```diff
@@ -0,0 +1,514 @@
...
+- Every deploy stone automatically becomes a **stronghold**.
+
+## Zones
+- **Black territory**: rows 1–9 · **White territory**: rows 11–19 · **Borderline**: row 10.
+- **Attack zone** (the *only* place that scores): the opponent's territory + the borderline.
+- **Defense zone**: your own territory + the borderline.
+
+## Scoring
+- **Territory** (empty points enclosed in the attack zone): **+2 / point**.
+- **Siege** (enemy groups with no liberties, in the attack zone): **+3 / captured stone**.
```
benjaminpjones, line 520: do these exist? would they not be captured during the game?

packages/shared/src/variants/border_war.ts
```diff
@@ -0,0 +1,514 @@
...
+    "Border War (War Horn: Borderline) v9.0\n Players deploy on either side of a central borderline; the first 4 moves build strongholds on your own territory. Score territory, sieges and captures in the opponent's half or on the borderline; capturing or sieging on your own side replenishes troops. Breaking an opponent's scored enclosure earns defense points.",
+  rulesDescription: `# Border War (War Horn: Borderline) v9.0
+
+A **19×19** Go-like territory variant. Players fight across a central **borderline**
+(row 10) on a 19-row board. The goal is not to enclose the most area, but to **score by
+attacking into the opponent's half**.
+
+## Goal
+Score more points than your opponent (White gets **+5 komi** at game end).
+Each side has a **piece limit of 90**.
```
benjaminpjones: nit: komi and pieceLimit are config-driven.  Would be good to note that these values are the defaults.

packages/shared/src/variants/border_war.ts
```diff
@@ -0,0 +1,514 @@
...
+## Scoring
+- **Territory** (empty points enclosed in the attack zone): **+2 / point**.
+- **Siege** (enemy groups with no liberties, in the attack zone): **+3 / captured stone**.
+- **Capture** in the attack zone: **+4 / captured stone** (ordinary stones).
+- **Stronghold captured**: **+10**, no war damage, and no capture score.
+- **War damage**: **−1** for each ordinary stone captured from you.
+- **Komi**: White **+5** at game end.
+
+## Troop replenishment (no scoring)
+- Capturing or sieging enemy stones **on your own territory or the borderline**
```
benjaminpjones, line 527: tested this, and it appears captures *are* counted, likely due to this line:

```
normalLost[oppPlayer] += 1
```

would be worth a unit test.

packages/shared/src/variants/__tests__/border_war.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    // 布局结束，白在边境 bj(x1,y9) 放普通子；黑填其四气 aj(0,9)、cj(2,9)、bk(1,10)、bi(1,8)
+    g.playMove(0, "aj");
+    g.playMove(1, "bj"); // 白子下在边境
+    g.playMove(0, "cj");
+    g.playMove(1, "pass");
+    g.playMove(0, "bk");
+    g.playMove(1, "pass");
+    g.playMove(0, "bi"); // 提掉白 bj
+    const st = g.exportState({ phase: "play" });
+    expect(st.score.territory[0]).toBe(4); // +4 吃子分
```
benjaminpjones: these tests exercise the "live" score, but do not check the result of `finalizeScore`.  Recommend asserting the value of `g.result` as well.

packages/shared/src/variants/border_war.ts
```diff
@@ -0,0 +1,514 @@
...
+  last_move: string;
+  captures: NumP;
+  // 边境线专属
+  phase_hand: number; // 当前第几手（1基）
+  strongholds: Record<number, string[]>; // 双方据点（SGF 坐标字符串）
+  pieces_left: NumP; // 每方剩余兵力
+  deploy_done: boolean; // 布局阶段是否结束
+  // 实时分数面板（不含白贴目；点差由 finalize 结算）
+  score: {
+    territory: NumP; // 占领分：围空+2 + 围困+3 + 吃子+4/子（攻击区）
```
benjaminpjones: nit: I think the comment (and variable name) here is wrong. `score.territory` is only populated with the `annihilate` values (i.e. captures) in `liveScore()`.


```suggestion
    captures: NumP; // 吃子分：+4/子（攻击区）
```

packages/shared/src/variants/border_war.ts
```diff
@@ -0,0 +1,514 @@
...
+        ),
+        1: Math.max(
+          0,
+          this.config.pieceLimit -
+            this.pieces_on_board[1] +
+            this.soldierBonus[1],
+        ),
+      },
+      deploy_done: this.deploy_done,
+      score: this.liveScore(),
```
benjaminpjones, line 304: Are there intentions to expose this state to the user?  As far as I can tell, `liveScore()` is not used except for testing.

longgege-cd:
@pr560_reply.md

packages/shared/src/variants/border_war.ts
longgege-cd: Fixed: the constructor now throws for any board that isn't exactly 19x19, so an invalid config never creates a game.

packages/shared/src/variants/border_war.ts
longgege-cd: Fixed: computeTerritory now returns early when the outer border is empty, so an empty board is no longer credited to White.

packages/shared/src/variants/border_war.ts
longgege-cd: Good catch - removed that line in the config-validation commit.

packages/shared/src/variants/border_war.ts
longgege-cd: Implemented: the variant now uses a positional superko (SuperKoDetector, matching Baduk) plus a suicide check, which covers the basic one-point ko. Added a ko test and documented it in the rules.

packages/shared/src/variants/border_war.ts
longgege-cd: Done - renamed score.territory to score.captures (it holds capture points), updated the comment and tests.

packages/shared/src/variants/border_war.ts
longgege-cd: Done - the rule description now notes that komi and pieceLimit are config-driven defaults.

packages/shared/src/variants/border_war.ts
longgege-cd: Confirmed - captures do count toward war damage. The existing test already asserts casualty[1] == -1; I also added a finalizeScore/result assertion.

packages/shared/src/variants/__tests__/border_war.test.ts
longgege-cd: Fixed - added a test that asserts finalizeScore produces a valid g.result and numeric_result.

packages/shared/src/variants/border_war.ts
longgege-cd: Under standard Go keep-on-no-liberties there is no on-board sieged group, so that siege-capture term does not normally trigger here; troop replenishment from your own side is handled via the defense-zone capture path (covered by tests).

packages/shared/src/variants/border_war.ts
longgege-cd: It is used by the client for a live score panel (display only); the authoritative result comes from finalizeScore.

longgege-cd:
Good call - I've added a ConfigForm for the Border War variant in this branch, so game creation now shows a GUI instead of raw JSON. The board is fixed to 19x19 (rendered read-only), while Komi and piece limit are live-editable inputs. It's registered in `config_form_map` under the 'border war' variant, so it loads automatically when creating a game.

packages/shared/src/variants/border_war.ts
benjaminpjones: that's what I would expect it to be used for, but I did not see the liveScore values displayed anywhere

packages/shared/src/variants/border_war.ts
benjaminpjones: okay, so if siege capture does not trigger, is it dead code? can it be removed?

benjaminpjones (approved):

benjaminpjones:
Thank you for adding this variant, and addressing the feedback!

longgege-cd:
Thank you! This is the base version, and I would welcome any feedback on it. A Dawn of the Final Battle (决战黎明) expansion is planned, which will introduce a skill system.

PR #561, opened 2026-09-05
Move notifications count tracking into the store

benjaminpjones:
`NotificationsNav` tracked the unread count itself: it fetched `/notifications/count` on load and subscribed to the server's `NOTIFICATIONS_COUNT_EVENT` push. I'd like to centralize this in the store.

**Why?** I felt #559 touched too much data fetching, when it was supposed to be a simple layout change. I'll re-attempt once this is in.

**Related Bug-fix:** cf3db04 clears the count on logout. Previously it was only ever written while logged in, so the badge kept the previous session's number until a reload.

**Tested:** this PR introduces 4 tests on the store. Also manually ran the serve and check notifications still work

https://claude.ai/code/session_01K1iyNVQ1NGayfrb2DQ5VJa


merowin (approved):
this is a nice improvement, thanks!

PR #569, opened 2026-09-07
Deliver game notifications as browser push

benjaminpjones:
Adds Web Push  on top of the notification system we already have, so a subscribed game can reach someone with the site closed.

## Design

Push is a **second delivery channel for existing notifications**, not a parallel system. Every game notification already funnels through `addGameNotification`, but now we mirror notifications through push. There is a single new toggle on the Notifications page that determines whether to route notifications to browser push.

## Server

`packages/server/src/notifications/push.ts` wraps `web-push`. Subscriptions live in a new `push_subscriptions` collection keyed on the push endpoint, of which a browser has exactly one.

```
{
  userId: string;
  /** The push service URL, unique per browser profile. */
  endpoint: string;
  keys: { p256dh: string; auth: string };
  createdAt: Date;
};
```

Sending is best-effort and deliberately not awaited, so it does not block move submission.

Four routes under `/api/notifications/push/`:

- `GET key`: returns the VAPID public key needed to create a subscription from the browser
- `POST subscribe`: upserts the subscription record in `push_subscriptions`
- `POST unsubscribe`: deletes the endpoint's row
- `GET status`: returns whether the server has a subscription record in the database

## Client

- `renderNotification` moves from `NotificationsView.vue` into `shared`, so the notifications page and the push payload cannot drift apart in wording.
- We introduce a service worker for accepting push notifications - sw.js.
- `src/utils/push.ts` — permission, subscribe/unsubscribe, state. Notices when its stored subscription was made against a different VAPID key and re-subscribes, so a key rotation doesn't silently stop delivering.
- Toggle to enable push notifications on the notifications page.

## Deployment Note

**Push stays off until VAPID keys are set.** You can generate these with `webpush` or a number of other tools.

## Testing

39 new tests, manually tested on Firefox & Chrome.

Screenshot 2026-09-07 at 8 14 57 AM
Screenshot 2026-09-07 at 8 15 05 AM

---

https://claude.ai/code/session_01B56k2Vch1Qm5aseD22sv5u


benjaminpjones (commented):
Reviewed.

All files look like they belong:

- package.json,yarn.lock: `web-push` is a valid dep for this feature
- migrations: this is just adding indexes - one adds the uniqueness requirement
- server
    - notifications.ts: slots in the new method
    - push.ts: the bulk of the server-side logic
    - api.ts: 4 new endpoints
    - db.ts: we added a new collection for push subscriptions
    - users.ts: make sure to delete subscriptions with a deleted user
- shared:
    - notifications strings are used server-side too. Unclear what to do about i18n for these, but current status is untranslated
    - types for the API
- client
    - sw.js: service worker allows to handle incoming notifications when browser is not open
    - PushNotificationToggle - new UI element on notifications page
    - user.ts - unsub on logout
    - push.ts - the bulk of the logic to interact with the new endpoints
    - NotificationsView.vue - add the new button and move text to shared

Issues found and fixed during review:

- Removed overly-verbose README
- Removed trivial test w/ VAPID keys unset, made sure not to leak VAPID env across tests
- Change read-only `key` endpoint to `GET`
- logout fix

PR #574, opened 2026-09-12
Add a settings page

benjaminpjones:
This adds a `/settings` page and a user menu to reach it from.

The NavBar gets crowded with this settings page so it needed some organizing:

- on mobile: settings just gets added to the hamburger menu
- on desktop: we hide Profile, Settings, Logout (and Admin) under a username drop down. Log out is no longer a top-level navbar button.

## Settings

We aggregate a few existing settings that were scattered across the site:

- Language: this was not previously configurable - it was pulled from the browser
- Immediate submit: a localstorage option, currently shown on game page
- Push Notifications: currently shown on notifications page (not shown in screenshot - VAPID keys were unset)

We intentionally do not remove the settings from their current locations in this PR.  This is to keep it tightly scoped and prevent UI churn.

## For reviewers

- German translations could use eyes
- Do the Nav bar changes seem reasonable?

## Screenshots

Screenshot 2026-09-13 at 11 57 06 AM

Screenshot 2026-09-12 at 7 51 43 PM

Screenshot 2026-09-12 at 6 33 27 PM

https://claude.ai/code/session_01DxztD57nys9Yyv8LGjYKGK


packages/vue-client/src/locales/en.json
```diff
@@ -23,5 +25,11 @@
     "title": "Filter games",
     "all-variants": "All variants",
     "my-games": "games that I play"
+  },
+  "settings-page": {
+    "language": "Language",
+    "gameplay": "Gameplay",
+    "immediate-submit": "Immediate submit",
+    "immediate-submit-hint": "Send each move as soon as you place it, instead of confirming it first. Only applies to variants that support move previews."
```
merowin, line 33: > Only applies to variants that support move previews.

I think that in variants that don't support move preview, a move is always submitted immediately. So while the sentence is technically correct, it might cause more confusion than it solves.

German translation looks good to me.

benjaminpjones: Ah yeah that is weird. Will update

PR #577, opened 2026-09-19
Reject moves on finished games, and hide the buttons

benjaminpjones:
3 fixes here to prevent submitting moves after the game ends:

- **Server:** check `phase` instead of results - certain variants do not set `result`.
- **Client:** hide the special-move buttons and refuse to submit moves (board clicks and previews included) once the game is over.
    - To support this, we add `phase` to `GameStateResponse`

## Tested manually

The buttons are no longer shown:

Screenshot 2026-09-18 at 10 14 35 PM

Attempting to play a move results in a "game over" message:

Screenshot 2026-09-18 at 10 35 18 PM


packages/server/src/games.ts
```diff
@@ -260,6 +261,12 @@ export async function handleMoveAndTime(
   return game;
 }
 
+// A function rather than an inline check, so TypeScript doesn't narrow
+// `phase` to "play" for the rest of the caller after playMove() changes it.
```
benjaminpjones, line 265: I thought it's incredible TS can't detect that `game_obj` gets mutated **by its own method**, but this comment is true! discussion here: https://github.com/microsoft/TypeScript/issues/9998

